PR #2, opened 2016-02-01
Alphabetize imports

kennyt:
What is the alphabet?

From wikipedia: `An alphabet is a standard set of letters (basic written symbols or graphemes) which is used to write one or more languages based on the general principle that the letters represent phonemes (basic significant sounds) of the spoken language.`

Why do we want `import`s to be alphabetized?

Alphabetizing `import`s  can make parsing them at a quick glance of the eyes easier.


jfurrow:
hahaha.... thanks for the definition!

PR #16, opened 2016-06-03
Handle empty torrent list response.

jfurrow:
An exception was thrown when rTorrent returned an empty list of torrents.


gretel:
that went on quick.. regards

Issue #28, opened 2016-06-17
Connection using rpc socket


ArtyumX:
Hi, can you implement the connection using a rpc socket instead than using SCGI ?
Nice project btw, thanks :)

jfurrow:
I'll have to do more research on this as I'm honestly not familiar with this concept. I do see now that it's the recommended approach to RPC with rTorrent, documented here: https://github.com/rakshasa/rtorrent/wiki/RPC-Setup-XMLRPC

Thanks for filing the issue!

jfurrow:
Fixed in #28, props to @zp!

jockef3000:
Can anyone describe how to use the rpc socket? How does the URL look that I can use?

PR #35, opened 2016-06-18
Add scgi socket support

zp:
Enables use of a socket file to connect to rTorrent.

Implements #28


jfurrow:
Thanks! This works perfectly.

Issue #40, opened 2016-06-19
i18n support


mahoek:
I thought it would be a nice feature for Flood to support i18n.

_I've been frustrated by the old rutorrent interface since I began using it years ago. I've had the idea of creating something like u did here for a very long time. So when I found out about Flood last weekend I was really excited and quickly started testing it. It looks really nice._

_I am just getting started with React and started playing around with Flood this weekend. I've looked at the code for hours and I am beginning to understand how it was build more and more. I've learned a lot this weekend. Thanks for that!_

_What I did is trying to get some kind of i18n working in Flood. You can see what I did here:_

https://github.com/mahoek/flood/tree/feature/i18n

_Surely I did not build it like you would build this, but at least I wanted to mention it here. I am going to work it out further just to learn. If you have any tips or suggestions then I'd really like to hear them._

_It would be really cool if you'd (or someone else) make an IRC or Slack channel for Flood where we can talk about the future of Flood. What do you think?_

_Thanks and keep up the good work!_

jfurrow:
Thanks for this! I agree, this is absolutely necessary.

You've brought up a lot of really good questions, and unfortunately I don't have answers for all of them at this time. Let me take a bit of time to think before I say anything too binding :). One thing specifically I want to consider is what to consider `settings` vs `config`, and where/how these defaults should be defined/overridden. In the near future I would like to include a UI that allows full configuration of the application, which writes to the `config` file permanently.

As a whole, I think what you've done is great. I have a few arbitrary stylistic requests, which I should document somewhere for future contributions. But for now, here goes:
* I'd like to keep external module imports separate from internal module imports, where external module imports are all at the top of the file.
* When importing specific members of modules, let's avoid the whitespace after and before the brackets. So `import { foo } from 'bar';` should be `import {foo} from 'bar';`
* I would also like to alphabetize member imports: `import { foo, bar } from 'baz';` should be `import {bar, baz} from 'bar';`

So, given [this example](https://github.com/mahoek/flood/commit/e0d34833c6de8bc7521143a5b44da8d5224ffe9e#diff-f5ab0a2b30bf8f64f53ef9aa636c6c86R5), I would like for it to appear as:
```javascript
import {addLocaleData, IntlProvider} from 'react-intl';
import nl from 'react-intl/locale-data/nl';
import React from 'react';
import ReactDOM from 'react-dom';  

// ...internal module imports...
```

[...]

PR #42, opened 2016-06-20
i18n support

mahoek:
This is a WIP for i18n support for Flood.

- [x] AddTorrent modal
- [x] Auth settings & modal
- [x] Settings modal
- [x] Sidebar
- [x] Notifications
- [x] PriorityLevels
- [x] SortDropdown
- [x] TorrentDetail modal
- [x] TorrentList

As you have already noted in #40 loading/including the language should be improved. I will think about it.

Any suggestions for improvement would be really appreciated. I will keep working on it and keep this PR updated.


jfurrow:
This is looking great, thank you for spearheading this effort!

mahoek:
I am trying to rebase with the master before doing some more work but I am getting errors mainly because of the removed source/ folder. Does someone (@jfurrow) know how I can resolve this? I've been trying for a while now.

jfurrow:
I didn't think about the implication of that change... I'll look into a solution tomorrow and let you know... Sorry about that. 😞

jfurrow:
@mahoek I rebased this branch and pushed it here: https://github.com/jfurrow/flood/tree/feature/i18n. I did have to make one additional commit, but all of your commits are there. Let me know how you want to handle it from here, and if you'd like me to take any actions.

It seems to be working as expected!


Sorry again.

mahoek:
Thanks for your help again!

I am not really sure what I should do. If you open a PR am I able to merge it without any problems? Then that is fine by me. Or should I pull your branch and force push to my feature/i18n branch? Will this PR then update?

jfurrow:
Actually, I think it would be easier for you to pull the `feature/i18n` branch from my repo and force push to your forked repo's `feature/i18n` branch. That will update this PR and you can continue work as normal.

mahoek:
How do you think about how I get the translations in d9bf3fc and 9dce82e for the PriorityMeter and the SortDropdown? Is this ok?

I am also not sure if I should move all defaultMessages to `en.js` instead of placing them in the code itself. What do you think?

jfurrow:
I think it's fine how you've specified the messages for `PriorityMeter` and `SortDropdown`.

I am in favor of moving the `defaultMessage`s to `en.js`, but only if it's possible to fallback to `en` if a particular message is not defined in alternate languages. For example, if I create additional components and add messages in `en.js`, but they have not been defined in `nl.js`, what would `react-intl` return when trying to render the components for `nl`? I tried to quickly find an answer to this in the `react-intl` wiki but didn't find it at first glance.

mahoek:
I've read how the fallbacks work and I don't think it is the right way to store the `defaultMessage`s in the `en.js` file. The `IntlProvider` has `defaultLocale` but there is no `defaultMessages` or something like that. So it will never fall back to `en.js` as default. Please read [this](https://github.com/yahoo/react-intl/wiki/Components#intlprovider) section. Correct me if I am wrong.

I am now also struggling on how to handle the translations for the default `SortDropdown` value. The `displayName` for the default sort setting is set in the `SettingsStore` and the `displayName` is also saved in the `settings.db` when changed. Wouldn't it be better to save the sort state as a const or something. So we can just point to the const from the `SettingsStore` and the saved settings and handle the `displayName` in the `SortDropdown`.

Please let met know what you think about it.

jfurrow:
You're right about the `defaultMessage`s. So let's keep them as they are for now.

I see the problem with the way the `SortDropdown` is storing its values. I'm not sure I understand your suggestion — are you saying we store the `value` and `direction` of each option (e.g. `added`, `sizeBytes`, `uploadTotal` etc) in the database and then map each `value` to a `displayName` via a constant? I think that would be smart. I'd be happy to make this change, if you'd like, but I'm hesitant to create too many merge conflicts for you, so I've not done that yet. Let me know either way.

Thanks again for all of your work on this, I really appreciate it!

mahoek:
Yes, you're right. I think that it would be better to map the `value` to a `displayName`. Unfortunately, I have not that much time tonight. So if you have any spare time left it would be nice you could look at it. Or else I will look into it tomorrow evening.

Then (tomorrow) I want to do the notifications. I was thinking about how to do these. The messages for the notifications are set outside of the components. We probably need to do some kind of mapping here too or pass the [`MessageDescriptors`](https://github.com/yahoo/react-intl/wiki/API#message-descriptor) directly. Would like to hear your opinion on this one.

Also, do you know if I forgot to put any translations that have to be done on the todo list above? Or do you have any suggestions for things that should be added before we complete this PR?

I have some idea(s) but I don't know if we should do that in this PR cuz of the many changes and we don't want too many merge conflicts. **Idea:** Let the user change the language in the settings. So that every user can have its own language. The first user should also be able to select a language when creating his account.

You're welcome. I really enjoy contributing to this cool project and I am learning a lot by doing this.

Thank you for being so helpful.

jfurrow:
I've made the changes we discussed and merged them into the `master` branch, which unfortunately means you'll have a merge conflict in `SortDropdown.js` but it will be reasonably simple to resolve.

With this change, the only values stored in the database now will be the `property` (which I previously called `value`) and `direction` for sorting. The `property` is literally the property of the torrent to sort (`eta`, `name`, etc). Each property maps to a displayed name in `constants/SortProperties.js`. From here the translation could be handled similarly to the `PriorityLevels` that you've set up, unless you think this should be done differently.

As for the notifications, here are my thoughts...
* We should remove the keys `adverb`, `action`, and `subject` from the object passed to `NotificationStore.add` (and the corresponding props in the `Notifications/Notification` component).
* We should keep the `accumulation` and `id` keys (but should probably change the `id` keys to match the ID nomenclature you've set up in the translation files).
* The `Notifications/index` component will only pass the `id` and the `count` to the `Notification/Notification` component.
* The `Notification` component itself can make use of the [`FormattedMessage`](https://github.com/yahoo/react-intl/wiki/Components#formattedmessage) component with rich text formatting to achieve the green highlighted notification count.
* Maybe we can define the `defaultMessage` props in a constants file based on the notification ID.

As an example of the actual notification, here's some pseudocode to represent what I'm thinking:

In `constants/Notifications.js`:
```javascript
{
  'notification.add.torrent': 'Successfully added {countElement} {count, plural, ...}',
  'notification.remove.torrent': 'Successfully removed {countElement} {count, plural, ...}',
  ...
}
```

And in `Notifications/Notification.js`, that might look something like this:
```jsx
<FormattedMessage
  defaultMessage={Notifications[this.props.id]}
  id={this.props.id}
  values={{
    count: this.props.count,
    countElement: <span className="notification__count">{this.props.count}</span>
  }}
/>
```

I probably missed some important steps, but I think that would keep it pretty simple. Let me know what you think and if you want any help implementing this!

jfurrow:
Regarding the settings, let's handle that separately from this PR. I like the idea of the user selecting their language in the login screen, as well as a tab in the settings modal. I'd be happy to implement that as soon as this PR merges.

mahoek:
I have rebased onto master and got the translations for the SortDropdown working. I am not really sure if how I did this is a good solution cuz I store both the `id` and the `defaultMessage` in `../constants/SortProperties.js`.

Also got the notifications translated. **YAY!** 😄 I think I did it mostly how you've suggested above.

I think we got most of the translations done now. I want to re-check all the Dutch translations and make sure they are good because I think some of them can be translated better. If you want I can make a new PR for that later.

jfurrow:
Great work, @mahoek! I think this is great. I can't think of a better solution for the `SortProperties` constant so let's continue with this, it can always be adjusted later.

I'm going to go ahead and merge this. I'll work on storing the language selection in the database (as well as presenting the option at account creation). I'll also handle the overflowing text in the sidebar ("speed limits"/"settings"). Then on to other features :)

Really nice work here!

Issue #76, opened 2016-07-10
Improve torrent-detail/general tab


mahoek:
I was thinking about improving the torrent-detail/general tab.
```
General
Added:              July 10, 2016 11:07 AM
Location:           path/to/download
Scheduler:          Obeyed/ignored
Tags:               <tag1> <tag2> <tag3>

Transfer
Downloaded:         1% (630KB/63MB)
Peers:              0 of 2 connected
Seeds:              0 of 233 connected

Torrent
Size:               63MB
Hash:               883C6F02FC46188AC17EA49C13C3E9D97413A5A2
Creation date:      June 04, 2016 6:58 PM
Comment:            <torrent comment>
Type:               Public/private

Tracker
Tracker message:    <tracker message>
```
I think that free disk space should not be in the torrent detail modal. Maybe the sidebar is a better place? 

And what about the tracker information? Should it be here or on the Tracker tabs?

Your thoughts and suggestions would be highly appreciated.

jfurrow:
We have discussed some of this in Slack, but I'll add my feedback here:
1. I agree that disk space is a more global statistic. However, rTorrent itself considers it a per-torrent statistic. I think this is because rTorrent is reporting the available space on the disk which contains that torrent's data (I think — maybe I'm wrong about this). So if we were to display free disk space in the sidebar, we would need to display the data for all disks on the host, which could be a significant number of disks. So I think this needs to be explored in mockups.
2. For the "tracker" details, the benefit to being displayed on the "details" tab is that it's one less click away from the user. But, because the "tracker message" is used to report errors (as far as I can tell), it would probably be a lot more useful if it were displayed as a tooltip on the error icon in the torrent list when an error is detected. So, for now, I'm leaning toward leaving it on the "details" tab. If you feel strongly that it should be on the "tracker" tab, then that's fine with me too!

mahoek:
1. In the meantime, while we haven't moved the free disk space to the sidebar, do you think we should let it in the general tab?
2. I agree. I'll let it on the general tab for now. We can move that later.

jfurrow:
1. Yes, I do think it should stay in this tab for now.
2. Cool :)

jfurrow:
I'm considering this closed for now, great work!

PR #77, opened 2016-07-10
Feature/improve torrent detail general tab

mahoek:
WIP to improve the torrent-detail general tab as discussed in #76.

- [x] Style tags

screen shot 2016-07-10 at 14 46 24


jfurrow:
This is a great start!

I'm not sure about the right-aligned tabs on the details tab. What do you think about something like this?

client/scripts/components/Modals/TorrentDetailsModal/TorrentGeneralInfo.js
```diff
@@ -1,110 +1,132 @@
...
 import classNames from 'classnames';
 import React from 'react';
 
 import format from '../../../util/formatData';
 
 class TorrentGeneralInfo extends React.Component {
+  getTags(tags) {
+    return tags.map((tag, index) => {
+      return (
+        <span>{tag}</span>
```
jfurrow: We'll need to use the `key` prop on this tag, so: `<span key={index}>`

client/scripts/components/Modals/TorrentDetailsModal/TorrentGeneralInfo.js
```diff
@@ -82,7 +82,7 @@ class TorrentGeneralInfo extends React.Component {
         <li className="torrent-details__detail torrent-details__detail--tags">
           <span className="torrent-details__detail__label">Tags</span>
           <span className="torrent-details__detail__value">
-            {(torrent.tags.length == 0 ? this.getTags(torrent.tags) : <ValueNotAvailable />)}
+            {(torrent.tags.length ? this.getTags(torrent.tags) : <ValueNotAvailable />)}
```
mahoek: `torrent.tags.length` always returns 1 because `"".split(",").length` is 1. What do you think is the smartest way to deal with this?

mahoek:
Here is a screenshot with its current state.

screen shot 2016-07-12 at 21 31 41

Still not sure about what to use. Tables or this flexbox solution.

client/scripts/components/Modals/TorrentDetailsModal/TorrentGeneralInfo.js
jfurrow: Good catch! I fixed it in the model here: https://github.com/jfurrow/flood/pull/80 so you should be able to rely on the length being 0.

jfurrow:
This looks good! I think a `table` will give us the best behavior (with the columns resizing consistently).

mahoek:
If there is anything I should add/fix/change please let me know.

jfurrow:
Very nice! Thanks @mahoek!

PR #82, opened 2016-07-14
Update deprecated rTorrent commands

zp:
Allows Flood to function properly if the user has deprecated commands disabled in rTorrent.

Noticed this when attempting to run Flood with rTorrent built from the master branch, which has deprecated commands disabled by default.

### DIFF
--- a/server/models/ClientRequest.js
+++ b/server/models/ClientRequest.js
@@ -164,7 +164,7 @@ class ClientRequest {
 
   createDirectoryMethodCall(options) {
     this.requests.push(
-      this.getMethodCall('execute', ['mkdir', '-p', options.path])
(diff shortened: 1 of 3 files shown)


jfurrow:
Thanks for doing this! Do you know by chance if there exists a list of current commands? Or maybe I should be going off of the response from the `system.listMethods` command?

I'm embarrassed that you hunted down where all of these commands are defined as I've been meaning to organize these in some logical fashion...

zp:
https://github.com/rakshasa/rtorrent/wiki/rTorrent-0.9-Comprehensive-Command-list-(WIP)

Not sure if it has everything, I used this and double checking the rTorrent source.

jfurrow:
Thanks, that's extremely helpful.

PR #83, opened 2016-07-14
Allow deleting data when removing torrent

jfurrow:


zp:
Would it be possible to also support deleting data when removing torrents with the context menu?

jfurrow:
Absolutely! Done.

PR #95, opened 2016-08-07
Add support for long TLDs

stephdewit:
Some trackers use top-level domains with more than 6 characters.

At the moment, the longest non-IDN TLD is 18-characters long ([according to IANA](http://data.iana.org/TLD/tlds-alpha-by-domain.txt)).


jfurrow:
Nice, thanks!

PR #96, opened 2016-08-12
Fix torrents with more than one tracker

stephdewit:
Only the first tracker was usable in the sidebar filter.

'getCalculatedTrackers' method returned only the first domain.

Global regEx matching should not be used on a shared regEx instance. Look [here](http://stackoverflow.com/a/2142007).

### DIFF
--- a/shared/util/regEx.js
+++ b/shared/util/regEx.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const regEx = {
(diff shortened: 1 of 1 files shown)


jfurrow:
I think you forgot a commit here, I don't see anything that adjusts the `getCalculatedTrackers` method.

The regEx change makes sense.

stephdewit:
Hello.

My tests seem to show that the regEx modification is the only change required to fix the bug. Isn't it the case for you ?

For the record, if you set the 'g' flag, the lastIndex property of the regEx isn't reset between calls and the second 'exec' call doesn't match.

Let me now if there is something I don't understand.

jfurrow:
I see now, this is great! I was assuming there was a bug in `#getCalculatedTrackers`. Thanks!

PR #113, opened 2016-08-14
Improve unknown dates handling

stephdewit:
For some torrents, _rtorrent_ returns zero or an empty string in the `added` and/or `creationDate` properties. So it's displayed as '1/1/1970' in the UI.

capture d ecran 2016-08-14 a 00 09 24

capture d ecran 2016-08-14 a 00 11 32

I made a change to instead display a dash in the torrent list and _None_ in the detailed view.

capture d ecran 2016-08-14 a 00 18 05

capture d ecran 2016-08-14 a 00 20 40

I don't have much inspiration, do you think it's good enough as UX? If you think of something better, I'd be happy to improve this.

EDIT: I also fixed the missing calendar icon in the torrent list.

capture d ecran 2016-08-14 a 13 07 55

### DIFF
--- a/client/scripts/components/TorrentList/Torrent.js
+++ b/client/scripts/components/TorrentList/Torrent.js
@@ -105,9 +105,12 @@ export default class Torrent extends React.Component {
   render() {
     let torrent = this.props.torrent;
 
-    let added = new Date(torrent.added * 1000);
-    let addedString = (added.getMonth() + 1) + '/' + added.getDate() + '/' +
-      added.getFullYear();
+    let addedString = '\u2014'; // em dash
+    if (torrent.added) {
+      let added = new Date(torrent.added * 1000);
+      addedString = (added.getMonth() + 1) + '/' + added.getDate() + '/' +
+        added.getFullYear();
+    }
     let completed = format.data(torrent.bytesDone);
     let downloadRate = format.data(torrent.downloadRate, '/s');
     let downloadTotal = format.data(torrent.downloadTotal);
--- a/client/scripts/components/TorrentList/Torrent.js
+++ b/client/scripts/components/TorrentList/Torrent.js
@@ -21,6 +21,7 @@ const ICONS = {
   disk: <DiskIcon />,
   downloadThick: <DownloadThickIcon />,
   information: <InformationIcon />,
+  calendar: <CalendarIcon />,
   peers: <PeersIcon />,
   ratio: <RatioIcon />,
   seeds: <SeedsIcon />,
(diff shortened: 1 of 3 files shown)


jfurrow:
Good idea! The `added` date is a custom field that gets set when you add a torrent via Flood, rTorrent doesn't do this out of the box. I'm using the same custom field that ruTorrent sets, so Flood will display dates on torrents added via ruTorrent also.

In the torrent list, let's move the date to the end of the list of tertiary details. If the date doesn't exist, let's just not render it. So I'd suggest putting all of the tertiary detail `<li>` nodes in an array and then appending the date added to the end of the array. You can look at the `secondaryDetails` array in the same file as an example.

stephdewit:
Thanks for your advice! I did that change.

client/scripts/components/TorrentList/Torrent.js
```diff
@@ -147,6 +145,51 @@ export default class Torrent extends React.Component {
...
+    if (torrent.added) {
+      let added = new Date(torrent.added * 1000);
+      let addedString = (added.getMonth() + 1) + '/' + added.getDate() + '/' + added.getFullYear();
+
+      tertiaryDetails.push(
+        <li className="torrent__details--added">
+          <span className="torrent__details__icon">{ICONS.calendar}</span>
+          {addedString}
+        </li>
+      )
```
jfurrow: Let's add a semi-colon after this closing parenthesis.

client/scripts/components/TorrentList/Torrent.js
```diff
@@ -147,6 +145,51 @@ export default class Torrent extends React.Component {
       );
     }
 
+    let tertiaryDetails = [
+      <li className="torrent__details--completed">
```
jfurrow: We need to add a `key` attribute to each of these `<li>` elements because they are children of an array. A simple string will be fine, it just needs to be unique for each item in the array. Same for the pushed element on line 185.

jfurrow:
This is great! Two minor comments and then let's :ship:.

stephdewit:
Thanks for the review! 👍

I miss the semi-colon, oops.

I'm really new to React, and I didn't check the console for the key error, my bad!

server/models/Torrent.js
```diff
@@ -256,6 +256,28 @@ class Torrent {
     return trackerDomains;
   }
 
+  getCalculatedAdded(clientData) {
```
stephdewit: I see you removed the `Calculated` part, be careful when merging this.

jfurrow: Good call! Thanks for noticing that :)

jfurrow:
@stephdewit No problem, thanks for fixing these minor things!

Issue #79, opened 2016-07-11
Is there a docker image for flood?


serut:
Hello there !
I'm using docker on my VPS and I was wondering if you have planned to release an docker image or if it already exists. 
By the way, I don't see on your readme if flood is multi-user or not. It's a lot better to store specific configuration files for each user by just using the authentification with nginx & htaccess - instead of creating 1 VM per user.

jfurrow:
I do not have a docker image created (though I agree it would make sense). If you'd be willing to submit a PR for it, I'd gladly accept it!

Flood has no concept of user permissions by any means. Right now you can create multiple user accounts, but they all have the same permission — either full access to the Flood API, or none. I do plan to add role-specific permissions in the future, but to be honest I consider it to be a very low priority in comparison to other core features.

I'm not entirely sure what you're talking about by storing specific configuration files for each user and authenticating with nginx and htaccess. Could you expand on this?

serut:
> If you'd be willing to submit a PR for it, I'd gladly accept it!  

I can create a working docker image, but I'm not experienced with `baseimage-docker` (in order to create the cleanest image possible).  

I never run Flood, but if you create 2 users they will share the same XMLRPC entry point no ? So if I'm correct, that means they will see the same torrent list on the HMI.

> I'm not entirely sure what you're talking about by storing specific configuration files for each user and authenticating with nginx and htaccess.

In my case, each user has his own rtorrent process with his own XMLRPC entry point. When an user opens the webpage, nginx (or apache) asks for a user / password. The user enters one of the declared user of the .htpasswd file and he gets connected. From that, rutorrent gets the username and loads the config file for that user `/path/to/rutorrent/conf/users/{username}/config.php`, here is an example :  
```
<?php
$topDirectory = '/downloads/torrents/user4/';
$scgi_host = "localhost";
$scgi_port = 5006;
$XMLRPCMountPoint = '/RPCUSER4';
```

That allows me to create one VM for all users of my server instead of 1 VM for each user.  


This is how it's done on ruTorrent : https://github.com/Novik/ruTorrent/blob/master/php/util.php#L11  
And their issue when they implemented multi user : https://github.com/Novik/ruTorrent/issues/136

mahoek:
I've been thinking about multi-user for a while. Let me explain what I was thinking about:

Technically (I think) you can run multiple users who have all their own torrents, tags & settings on one Flood instance and one rTorrent instance. Flood could for example link torrents to a user and only show the torrents added by this user.

[...]

PR #118, opened 2016-08-16
Update npm scripts

jfurrow:
Update the npm scripts as discussed in #79

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,8 +3,11 @@
   "version": "0.0.0",
   "private": true,
   "scripts": {
-    "livereload": "NODE_ENV='development' ./node_modules/.bin/gulp livereload",
-    "start": "npm prune && npm install && ./node_modules/.bin/gulp dist && ./node_modules/.bin/nodemon ./server/bin/www"
+    "preinstall": "npm prune",
+    "prestart": "./node_modules/.bin/gulp dist",
+    "start": "./node_modules/.bin/nodemon ./server/bin/www",
+    "start:production": "node ./server/bin/www",
+    "livereload": "NODE_ENV='development' ./node_modules/.bin/gulp livereload"
   },
   "dependencies": {
     "axios": "^0.12.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -3,8 +3,11 @@
   "version": "0.0.0",
   "private": true,
   "scripts": {
-    "livereload": "NODE_ENV='development' ./node_modules/.bin/gulp livereload",
-    "start": "npm prune && npm install && ./node_modules/.bin/gulp dist && ./node_modules/.bin/nodemon ./server/bin/www"
+    "preinstall": "npm prune",
+    "prestart": "./node_modules/.bin/gulp dist",
+    "start": "./node_modules/.bin/nodemon ./server/bin/www",
+    "start:production": "node ./server/bin/www",
+    "livereload": "NODE_ENV='development' ./node_modules/.bin/gulp livereload"
```
serut: start:watch is more commun

serut:
Pre-Requisites 
- add : if you are want to use docker, rtorrent needs to open the XMLRPC entry point to the network. Ensure that you put `0.0.0.0` in the config of `.rtorrent.rc`

Starting the Server 
- rename to Starting the Server on dev
- add 0. npm install
- change 1. command to `npm run start:watch`

Add Starting the server on production:
- npm install --production (no devDependencies)
- npm run start (no nodemon => no watcher)

Otherwise it's a good improvement ! :+1:

package.json
jfurrow: I like it. Done!

jfurrow:
@serut I structured the readme a little differently, but it should cover all of your concerns! Thanks for the suggestions.

serut:
You may want to use webpack because it can watch, reload, create the production build... Instead of gulp. Your readme is perfect by the way

jfurrow:
I've used Webpack's watch system in the past and wasn't pleased with its performance, but I'll check it out again in the future for sure.

serut:
Webpack saves files on RAM so its quite good. And when I said webpack instead of gulp maybe I made a mistake and webpack can uses gulp. Dont know for sure but there is a lot of functunality with webpack

jfurrow:
I'm actually using Webpack with Gulp already, just not making use of Webpack's watch feature as I found it slow in the past. It's certainly possible I didn't have it configured optimally, but I am very happy with the speed of the current setup.

PR #122, opened 2016-08-20
Fix missing dates in details

stephdewit:
Since the merge of #113, it was impossible to open the `TorrentGeneralInfo` view for torrents without an added or creation date.


jfurrow:
Yikes, sorry about that. Thanks for the fix.

PR #124, opened 2016-08-29
Add French Translation

saldru:
Add French translation for Flood Interface.

It miss some variables for translation. In French, we are not using the "_Byte_" as default unit but the "_octet_" unit. In English, you are using **B**, **kB**, **MB**, **GB**, **TB**.... In French, we use **o**, **ko**, **Mo**, **Go**, **To**...
The variables for the Feeds Menu are missing for translation.
Thanks,
Saldru


jfurrow:
Thanks very much! Regarding the unit abbreviations, that's a good point. They should be translated also; I'll add support for this.

PR #127, opened 2016-09-02
Complete i18n

stephdewit:
Hello.

I translated a few components which weren't yet.

I only checked the main view and the torrent details modal, not the `Settings` and `RSS` views. But it's a beginning. 😄

Sorry for the big PR, maybe I should have done multiple smaller ones. I got carried away! 🐵

PR #136, opened 2016-09-17
Move formatting into components

stephdewit:
I created a Ratio component, and moved the code from formatData.js into it.

So it's consistent with what have been done in #127 and all formatting is done within components.


jfurrow:
:tada: Thanks!!

PR #137, opened 2016-09-18
Display dependency checklist as application loads

jfurrow:


mahoek:
Nice work @jfurrow, looks great!

PR #140, opened 2016-09-20
Improve translations

mahoek:
I gave some translations better ids (I think) and added some missing Dutch translations.


jfurrow:
Thanks! 🎉

PR #144, opened 2016-09-28
Replace whitespaces in filenames

Joris33:
Whitespaces in .torrent files' names are causing issues with rTorrent (tested on 0.9.6).

### DIFF
--- a/server/models/client.js
+++ b/server/models/client.js
@@ -32,6 +32,8 @@ var client = {
     // filesize of 524 kilobytes or less. This allows the user to send many
     // torrent files reliably.
     files.forEach((file, index) => {
+      file.originalname = file.originalname.replace(/\s+/g, '.');
+      
       let fileRequest = new ClientRequest();
       fileRequest.add('addFiles', {files: file, path, start});
 


jfurrow:
Thanks for the PR! I'd like to test this a little bit before accepting.

jfurrow:
@Joris33 What specifically were the issues you noticed in rTorrent 0.9.6 with whitespaces in the file name? I've tested several different versions of rTorrent (including 0.9.6) and can't find any problems with whitespace.

The files are sent as raw data to rTorrent, so I don't think rTorrent even knows or cares what the filename is.

Joris33:
Torrents were stuck at 0% with the error 'Junk at end of input' about a start command that failed, with no more details (I don't have anything scheduled when a torrent starts tho).

If I manually removed the spaces from the .torrent file's name and tried sending it again, it was working. I tried multiple times with the same results.

Changing the name sent by Flood to rTorrent solved this.

rTorrent runs inside a Docker container based on Alpine Linux: https://hub.docker.com/r/xataz/rtorrent/ and I've been using Chrome.

jfurrow:
Thanks for the explanation! I will try to reproduce with the docker image but I'm also happy to accept the PR. Sorry for the delay.

PR #148, opened 2016-10-22
Notifications

jfurrow:
This PR:
* Introduces torrent notifications in the UI (currently only when a torrent has finished)
* Implements continuous torrent data polling on server


jfurrow:
For torrent completion notifications, Flood needs to check if rTorrent is checking the hash, otherwise false notifications will be delivered.

PR #170, opened 2016-11-05
ℹ️ Add Mediainfo support

jfurrow:
This PR adds `mediainfo` support in the torrent details modal. 

It is required that you have `mediainfo` installed and available in the PATH of the Flood server's user.


maximuml:
Mediainfo working only for folders, not for files

> An error occurred while running mediainfo on the server. Check that mediainfo is installed and available in the PATH to Flood.

PR #194, opened 2016-12-05
Add support for stopped status

Jellyfrog:
Not sure about the icon but looks decent enough for me


jfurrow:
Thanks! I may adjust the icon in the future, but it's not bad IMO!

PR #209, opened 2017-01-03
more install steps + precisions

noraj1337:


noraj1337:
Per user install do not easily allow to create a systemd service file. I'll create a "Install flood with a service account" page on the wiki.

PR #257, opened 2017-02-05
Look for configuration file relative to server

SamuelTilly:
This makes it possible to start the server from another location using for example `node /path/to/flood/server/bin/www`


jfurrow:
Good call @SamuelTilly and thanks for the PR!

PR #297, opened 2017-03-24
Fix flags alignment

wopian:
Flags in the peer list were centered 1px too far to the right, getting cut-off by `overflow: hidden` of its container - most noticable with Japan's flag.

Before:

After:

### DIFF
--- a/client/sass/components/_peers-list.scss
+++ b/client/sass/components/_peers-list.scss
@@ -11,7 +11,7 @@
 
     &__image {
       height: 10px;
-      left: 50%;
+      left: 49%;
       position: absolute;
       top: 0;
       transform: translateX(-50%);


jfurrow:
Sorry but I don't think this is the right solution.

Right now the set of flags vary by a few pixels in width (maybe this reflects the aspect ratio of actual flags? I haven't done much research on this). Flood displays all flags within a 15x10px box (width x height), centered horizontally. It sets their height to 10px but lets the browser handle the width, to preserve the aspect ratio of the original image, and cuts off anything outside of the 15px wide area. This results in sub-pixel discrepancies among the various flags, which is what you're noticing.

I think either of the following solutions would be best:
1. Find a set of flag images with consistent dimensions with an appropriate license.
2. Stretch the current set of flag images to fit within the 15x10 (or whatever dimensions are best) container, ignoring the original aspect ratio.

The first is the best, but I can't find a set that meets the requirements. So I think the second would be better.

What do you think @wopian?

wopian:
Would [`lipis/flag-icon-css`](https://github.com/lipis/flag-icon-css) work? They're a consistent 4:3 ratio and licensed under MIT

jfurrow:
That's perfect, @wopian! Thanks! I'll create an issue for this and close this PR.

PR #300, opened 2017-03-26
Changed from listening on all interfaces to listening on a specified one

Sanguinaris:
Enables integration into Webservers proxying the data so the nodejs server is not exposed to the public.


frebib:
From a security perspective, this really should be default (y)

jfurrow:
Really sorry for the terrible delay, I'll experiment with this tomorrow.

Thanks for your patience!

jfurrow:
Thanks @Sanguinaris, this looks good!

PR #304, opened 2017-04-08
Add support for Docker

bryanalves:
Including a configuration file that allows setting of most settings by
injecting environment variables into the container.


jfurrow:
Thanks!

PR #315, opened 2017-04-22
Link changelog to diff

wopian:
Links the version to its relevant changes diff page - its what the [] are 😉  

Also a good idea to roughly follow [this](http://keepachangelog.com/en/0.3.0/) to ease readability ([example](https://github.com/wopian/hibari/blob/master/CHANGELOG.md)) 👌


jfurrow:
Thanks @wopian!

PR #328, opened 2017-04-25
Option to download files directly to destination

### LINKED ISSUE #327: Option to download files directly to destination
If I'm downloading a torrent that is a folder, I'd like to have the files inside the folder placed directly in the selected destination.

The current functionality is such that if I make a torrent from this folder: `bar/baz.abc`
And I set the destination to `foo/`
the file will be available in `foo/bar/baz.abc`. I want the file to be placed in `foo/baz.abc`.

ruTorrent has an option on the "add torrent" dialog labeled "Don't add torrent's name to directory". You can see it in [this screenshot](https://raw.githubusercontent.com/wiki/Novik/ruTorrent/images/scr3_big.jpg)

nVitius:
This patch adds functionality for #327

### DIFF
--- a/server/models/ClientRequest.js
+++ b/server/models/ClientRequest.js
@@ -114,6 +114,7 @@ class ClientRequest {
   addFiles(options) {
     let files = this.getEnsuredArray(options.files);
     let path = options.path;
+    let useBasePath = options.useBasePath;
     let start = options.start;
     let tagsArr = options.tags;
 
@@ -123,7 +124,11 @@ class ClientRequest {
       let timeAdded = Math.floor(Date.now() / 1000);
 
       if (path && path !== '') {
-        parameters.push(`d.directory.set="${path}"`);
+        if (useBasePath) {
+          parameters.push(`d.directory_base.set="${path}"`);
+        } else {
+          parameters.push(`d.directory.set="${path}"`);
+        }
       }
 
       parameters = this.addTagsToRequest(tagsArr, parameters);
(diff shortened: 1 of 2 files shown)


nVitius:
So far, I've added and tested the API code for this. I still need to write the code for the frontend portion.

jfurrow:
Thanks @nVitius! As for the UI, I'm not 100% decided on a particular solution, but I think it's time to start hiding some of the options in this modal. So my initial thought is we move the "tags" textbox and "start torrent" checkbox into an "advanced settings" section that is hidden by default. See the following mockup:
flood-add-torrents-advanced

So, when revealed, the advanced settings would contain a textbox for tags, a checkbox to start torrents on load, and another for "don't append torrent name to path" or whatever is appropriate.

Alternatively this modal could be redesigned to accommodate displaying all of these options in a more organized manner. What do you think?

nVitius:
I don't think the modal is particularly complicated that we need to do that. I also don't think tags and start torrent really fit under "advanced settings". At least not by themselves. If the goal is to make the modal as straight-forward as possible, I feel like it would make more sense to put those in there if you put "destination" in there as well. Personally though, I use those options all the time.

jfurrow:
Yeah, that's a good point. Personally I use the destination textbox every single time.

I'll have to mock up a solution for this in which the fields are not hidden at all. Are you interested in implementing the UI? If not, that's fine, I'll merge this PR as-is and build the UI separately.

nVitius:
I've never really worked with React before, but I can give it a shot. We can merge this one now if you want, and I can work on the UI over the weekend.

jfurrow:
That sounds great! I'm happy to help in any way possible, feel free to ping me on Slack.

wopian:
Tags and the start toggle in advanced would be a PITA for me - I use them every time, whereas I never change the destination from the default 😅

PR #340, opened 2017-05-01
Base Path Checkbox in Destination

nVitius:
This patch adds UI functionality for #327


nVitius:
@jfurrow Which of these do you like better for the checkbox location?
one
two

The code I've pushed up is for the latter. I still have to tie in the checkbox logic.

wopian:
First 👌

jfurrow:
Thanks @nVitius! I prefer the second option, given that we center the textbox. I'd also like to remove the question mark, so that the checkbox is more of an affirmation rather than an answer to the question.

I realize the styles for forms isn't super flexible in terms of centering the checkbox, and I should work on this at some point.

I'm also wondering how this might affect the file deletion behavior. Is it possible for a multi-file torrent to be downloaded to a directory which contains other files that are not related to the torrent? If so, the file deletion logic will need to be adjusted.

nVitius:
@jfurrow I managed to get it to center, though I had to add a new style for this particular one.

Does Flood actually do the file deletion itself? I thought rtorrent would handle that. In that case though, we would definitely have to update the deletion logic to ensure that only that torrent's files get deleted. We could do this one of  two ways:
1. Fetch file list and delete those files individually
2. Have rtorrent move the files to a directory under `/tmp` and delete them there (does Flood do the actual moving too though? It wouldn't work in that case).

nVitius:
Also, I removed the question mark from the text.

jfurrow:
Flood handles all of the file deleting. As far as I know, rTorrent doesn't offer this functionality. Here's where Flood does it: https://github.com/jfurrow/flood/blob/master/server/models/client.js#L96-L117

Flood does the actual moving of files, too. So, I think option 1 is the best. It's unfortunate that we have to do the additional request for files, but it's the only way to be sure we don't unintentionally delete other files.

nVitius:
@jfurrow Given the [conversation](https://flood-talk.slack.com/archives/C1Q9X3N9W/p1494861890943130) we had on Slack the other day, it seems like this file deletion problem already exists. Do you think we could merge this in and address that in another issue? I haven't had much time to work on Flood since this PR, but would like to start doing so again next week.

jfurrow:
@nVitius I'm hesitant to merge this PR because I'm afraid that using this will cause user's data to be deleted. I think the deletion logic needs to be improved first before merging this. Sorry!

jfurrow:
@nVitius Now that #376 merged, let's get this one merged too! Can you please rebase this PR and then commit the compiled assets (by running `npm run build-assets`)?

nVitius:
@jfurrow assets updated

jfurrow:
Great work @nVitius!

PR #363, opened 2017-05-16
Made a quick logo and favicon

NoLooseEnds:
Made a quick logo and favicon so it’s faster to find in the browser. If you allready have a logo, feel free to use this temporary or not at all. :)


jfurrow:
Great work! Thanks so much for doing this.

PR #364, opened 2017-05-20
Flood @ <hostname> in the title

NoLooseEnds:
Added hostname to the title of the page so it’s easier to differentiate between servers.

I’m not experienced with node at all, so not sure if I did this “after the book”.

### DIFF
--- a/server/routes/main.js
+++ b/server/routes/main.js
@@ -3,12 +3,14 @@ let config = require('../../config');
 let express = require('express');
 
 let router = express.Router();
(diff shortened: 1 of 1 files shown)


jfurrow:
Sorry, but I do not think this is the correct solution. Let's maybe open an issue for the concern you are addressing, which is that you'd like the servers to be more easily identifiable when you have multiple tabs running.

I do appreciate that you submitted the PR, thank you for that!

PR #370, opened 2017-05-27
Add mediainfo for docker containers

Centzilius:
In order for mediainfo to work in docker containers we need to install it (obviously^^)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,5 @@
 FROM node:7.8.0-onbuild
-COPY config.docker.js config.js
-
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,5 +1,7 @@
 FROM node:7.8.0-onbuild
 COPY config.docker.js config.js
 
+RUN apt-get update && apt-get install -y mediainfo && rm -rf /var/lib/apt/lists/*
```
nVitius: we should run this before the COPY of the config file

jfurrow:
Thanks for the PR!

PR #372, opened 2017-05-29
Implement event stream to communicate rTorrent activity

jfurrow:
If anyone is looking at this, sorry the PR is so huge!

This PR removes all client-side API polling and implements an event stream subscription. All data related to the torrent list, transfer rate, torrent taxonomy, and notification count is provided via the event stream. This significantly reduces the amount of data consumed by the Flood client and interacting with torrents (e.g. starting, stopping, etc) should feel more responsive.

The Flood server still polls rTorrent on an interval, defined in `config.js` as `torrentClientPollInterval` and defaults to 2 seconds. If the data has changed since the last poll, then an event is emitted with relevant changes.

The only exception to this is when fetching a specific torrent's detailed data (peers, files, and tracker data); this should probably refactored as a separate event stream in the future.

Although there is still more work to do, the Flood server has been heavily refactored in this change. From this point forward, all rTorrent client requests should be performed via the `clientRequestService`.


noraj1337:
Wow, amazing

midzer:
Awesome!
I experienced massive CPU usage reduction by node process after pulling latest master with this change.

jfurrow:
@midzer That's great to hear, but honestly unexpected... I did refactor a significant portion of the Node server's logic, but essentially it's doing the same thing. Maybe there was some obfuscated performance hit that I inadvertently fixed?

PR #376, opened 2017-06-08
Only delete files belonging to torrent

jfurrow:
~~(Includes commits from https://github.com/jfurrow/flood/pull/372, which should merge first)~~

This PR will delete only the data that belongs to a specific torrent. With the previous file deletion logic, Flood would try to the torrent's directory if it were a multi-file torrent. This was a problem because rTorrent provides the option to omit the base path when adding a torrent, in which case Flood would delete the torrent's parent directory. You can probably see where this is going...

With this PR, any time a user requests to delete a torrent's data, Flood will first get the list of files from rTorrent, then delete only those files. It should be impossible to delete files which don't belong to the torrent, unless users move files into a torrent's nested directory...

### DIFF
--- a/server/constants/clientRequestServiceEvents.js
+++ b/server/constants/clientRequestServiceEvents.js
@@ -7,6 +7,7 @@ const clientRequestServiceEvents = [
   'PROCESS_TORRENT_LIST_END',
   'PROCESS_TORRENT_LIST_START',
   'PROCESS_TRANSFER_RATE_START',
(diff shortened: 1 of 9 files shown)


midzer:
Just deleted some torrents with flood on master and realized folder from those torrents not being deleted any more.

Maybe it belongs to this PR.

jfurrow:
@midzer Created an issue for that https://github.com/jfurrow/flood/issues/399

PR #409, opened 2017-07-24
Ignore temp server files

bengadbois:


.gitignore
```diff
@@ -4,3 +4,4 @@ node_modules
 npm-debug.log
 server/db
 server/assets/*.map
+server/temp
```
Soullivaneuh: Should be `/server/temp`.

Otherwise, it will ignore every `server/temp` files/folder on the project.

It should be the same for the rest of the file BTW.

jfurrow: You're right, using full paths would be safer.

jfurrow: https://github.com/jfurrow/flood/pull/418

PR #410, opened 2017-07-25
Add HTTP2 support to server

bengadbois:
I'm not sure if this is the best way to do it, basically just switched out the express server. It loads everything (html, js, all the assets) as http2.

Page loads don't _feel_ any faster, but the header compression and http2 server push is happening, so it should be slightly faster on slow connections.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "pug": "^2.0.0-beta3",
     "rimraf": "^2.6.1",
     "serve-favicon": "^2.3.0",
+    "spdy": "^3.4.7",
     "xmlrpc": "^1.3.0"
   },
   "devDependencies": {
--- a/server/bin/www
+++ b/server/bin/www
@@ -32,7 +32,7 @@ if (config.ssl) {
     return;
   }
 
-  server = require('https').createServer({
+  server = require('spdy').createServer({
     key: fs.readFileSync(config.sslKey),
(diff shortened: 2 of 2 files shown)


jfurrow:
Thanks for the PR!

Let's make this configurable by adding an option in `config.template.js` (something like `useHTTP2`). Then we can default to `https` but use `spdy` if the configuration is present.

bengadbois:
@jfurrow is there any reason someone would use `https` over `spdy`? The `spdy` module will gracefully fallback to HTTPS 1.1 in the cases that HTTP/2 or SPDY isn't supported by a browser (or something proxying in front).

jfurrow:
@bengadbois Oh cool, I didn't know that. I honestly don't know much about HTTP2 or SPDY.

Anyway, the package seems well supported, but this isn't working for me... the server just never responds. Do I need to generate the certificate in a special way?

bengadbois:
I didn't need to, just an npm install and flood restart. So the flood server never responds to any HTTPS requests then (HTTP/2 only works with HTTPS)? Are you hitting it directly or proxying with something?

For me the same cert file, same port, nothing changed `config.json`, all worked the first time.

jfurrow:
Alright, yeah I tried this again and it's working as expected, I'm not sure what happened earlier but let's ship it 🚢 ! Thanks for the PR

PR #417, opened 2017-08-04
Improve development & build process

jfurrow:
This PR uses the development & build tools from [create-react-app](https://github.com/facebookincubator/create-react-app). It provides a much nicer development experience and a simpler build process than Flood had before.

The biggest impact to users of Flood is that you'll have to build static assets locally on your machine. I'm sure some people will be upset about this, but it's for the best. ❤️

I'm considering this a WIP until I have time to test it more thoroughly.

Things to note:
* Users need to build their own assets by running `npm run build`. I foresee some problems with dev dependencies not installing properly, which sucks, but we'll work through it. A good Dockerfile would fix this.
* npm scripts have changed. The old ones still exist, but display a deprecation warning.
* npm version 5+ should be used to take advantage of npm's new `package-lock.json`.

Still to do...
- [x] Bring back `favicon` & `manifest.json`, I lost them along the way :(
- [x] Test some things
- [x] Look for unused & outdated dependencies


jfurrow:
Finally going to pull the trigger on this...

PR #430, opened 2017-08-13
Use process.exit() instead of graceful return

SanPilot:
ESLint also raises this warning.


jfurrow:
Thanks for the PR! This seems reasonable, but I'll need to test it first.

PR #441, opened 2017-09-05
Fix #401

### LINKED ISSUE #401: Mediainfo error
Getting this when try to mediainfo a torrent file in the ui

An error occurred while running mediainfo on the server. Check that mediainfo is installed and available in the PATH to Flood.

Shows this 

GET /api/mediainfo?hash=65F878981993A86211C13EAA0A2D11C4E9098C67 500 161.690 ms 

Running Debian 8

SavageCore:
`client.getTorrent()` was removed in 7f1acd711bb0a69b878af655f90587a1988c90db so use new torrentService

### DIFF
--- a/server/util/mediainfo.js
+++ b/server/util/mediainfo.js
@@ -2,7 +2,7 @@
 
 let child_process = require('child_process');
 
-let client = require('../models/client');
+const torrentService = require('../services/torrentService');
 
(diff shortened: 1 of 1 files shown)


jfurrow:
Oh, that's an embarrassing bug — thanks for looking into it and fixing!

PR #442, opened 2017-09-10
Fix RSS parsing

### LINKED ISSUE #433: RSS Feeds are parsed incorrectly
Currently, Flood uses the URL from the `<link>` tag of an RSS item to download. However, most torrent sites' RSS feeds use the `url` attribute of the `<enclosure>` to contain the actual torrent file/magnet link. Often the `<link>` tag links to the listing page for the torrent.

See the RSS feed for YIFY, eztv, etc. 

The FeedSub module's documentation is lacking, and I couldn't get it to work in a node shell, but the correct url must be grabbed from the attribute.

SanPilot:
Re-implemented functionality from FeedSub module because it didn't support attributes. I have tested changes but recommend further testing. Fixes #433.


jfurrow:
Thanks @SanPilot! I'll take a look at this ASAP.

jfurrow:
Well, this works! There are some things I'd like to change, but since I waited so long to review this, I'm going to accept it as-is and amend afterwards.

Thanks @SanPilot!

jfurrow:
@SanPilot I merged this a little too early; it actually broke some basic functionality, so I'm going to revert for now. Sorry!

SanPilot:
@jfurrow No need to be sorry. I don't want to make Flood unstable. The issue is that the RSS feature is very difficult to test. And, in retrospect, the changes were essentially a hack.

I plan on smoothing out the rough areas. Can you tell me exactly which things broke so I can focus on those first? Thanks.

PR #451, opened 2017-09-21
Drag & drop torrent to torrentList

### LINKED ISSUE #350: [Request] Drag & drop torrent to torrentList
Adding a torrent file is currently a many-step process (click "Add torrent" → click "By file" → drag & drop → click "Add file"). Since adding a single torrent is done quite often, it would be nice to be able to drag & drop to add and start the torrent, as soon as the site has loaded.

D3X:
cursor

Closes #350

### DIFF
--- a/client/javascript/components/torrent-list/TorrentList.js
+++ b/client/javascript/components/torrent-list/TorrentList.js
@@ -2,12 +2,14 @@ import {defineMessages, formatMessage, FormattedMessage, injectIntl} from 'react
 import _ from 'lodash';
 import classNames from 'classnames';
 import CSSTransitionGroup from 'react-addons-css-transition-group';
+import Dropzone from 'react-dropzone';
 import React from 'react';
 import ReactDOM from 'react-dom';
 
 import ContextMenu from '../general/ContextMenu';
 import CustomScrollbars from '../general/CustomScrollbars';
 import EventTypes from '../../constants/EventTypes';
+import Files from '../icons/Files';
 import ListViewport from '../general/ListViewport';
 import LoadingIndicator from '../general/LoadingIndicator';
 import PriorityLevels from '../../constants/PriorityLevels';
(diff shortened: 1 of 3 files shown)


jfurrow:
That looks awesome! Thanks for the PR! I'll review this ASAP.

jfurrow:
This is almost there, but I think it could use a little more polish.

If the browser isn't in focus when I drag a file onto the torrent list, it doesn't display the dropzone graphic until I drop the file (this is presumably when the browser receives focus). Also, the transition value is a little slow — maybe let's try `250ms`. Lastly, if I drop a directory on the window, it throws an error but the UI doesn't respond at all (I'd be fine with letting this one slide for now).
sep-21-2017 19-57-47

For reference, the "Add Files" modal responds to a dragged file even when the browser isn't in focus:
sep-21-2017 20-02-57

D3X:
@jfurrow I've improved the responsiveness by switching to a pure-css solution.

As for drag-n-dropping directories - the "add torrent" modal does not really support that either. It adds the directory to the list, but uploading it fails (obviously). I think there should be a global solution for this issue, not one specific to this PR/ticket.

larsjohnsen:
Thanks, @D3X. I tested this, and it works as desired in issue #350.

jfurrow:
> As for drag-n-dropping directories - the "add torrent" modal does not really support that either. It adds the directory to the list, but uploading it fails (obviously). I think there should be a global solution for this issue, not one specific to this PR/ticket.

You're right — thanks for pointing this out.

> Additional comment: I don't think the transition is necessary. It would be nice to have instantaneous feedback that dropping the torrent is accepted.

This is a valid point, but I think a small delay is acceptable. Right now Flood has 250ms transitions pretty much everywhere, which is too long of a time. This is another thing that should be addressed in a different PR, I think.

jfurrow:
Thanks @D3X — I hope my negligence in the review of these PRs hasn't discouraged you from contributing too much. You're doing great stuff and this is a huge help. <3

D3X:
@jfurrow no worries, neither of us works on this full time :)

Issue #415, opened 2017-08-02
Speed state on the browser title


Soullivaneuh:
It would be great to have the speed state on the browser title.

For example: `Flood - D: 250 kB/s - U: 100 kB/s`

Or even shorter, but less clear: `Flood - 250 kB/s / 100 kB/s`

PR #452, opened 2017-09-21
Download/Upload rates in the window title.

D3X:
_1_79kb_s_ _106kb_s___flood

Closes: #415

### DIFF
--- a/client/javascript/components/general/WindowTitle.js
+++ b/client/javascript/components/general/WindowTitle.js
@@ -4,22 +4,11 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 
 import * as i18n from '../../i18n/languages';
+import {compute, getTranslationString} from '../../util/size';
 import EventTypes from '../../constants/EventTypes';
-import Size from './Size';
 import TransferDataStore from '../../stores/TransferDataStore';
 
 const METHODS_TO_BIND = ['handleTransferChange'];
-
-
-let Title = ({down, up}) => (
-  <span>
-    &darr; <Size value={down} isSpeed />{' '}
-    &uarr; <Size value={up} isSpeed /> |{' '}
-    Flood
-  </span>
-);
-
-Title = injectIntl(Title);
 
 
 class WindowTitle extends React.Component {
@@ -29,8 +18,6 @@ class WindowTitle extends React.Component {
     this.state = {
       title: 'Flood',
     };
-
-    this.title = document.createElement('span');
 
     METHODS_TO_BIND.forEach((method) => {
       this[method] = this[method].bind(this);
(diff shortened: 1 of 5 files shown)


client/javascript/components/general/WindowTitle.js
```diff
@@ -0,0 +1,85 @@
...
+      <IntlProvider locale={intl.locale} messages={i18n[intl.locale]}>
+        <Title down={summary.downRate} up={summary.upRate} />
+      </IntlProvider>
+    ), this.title);
+
+    this.setState({
+      title: this.title.innerText,
+    });
+
+    ReactDOM.unmountComponentAtNode(this.title);
```
jfurrow: I'm assuming you chose to create this temporary DOM node and grab its `innerText` to leverage the `Size` component, but unfortunately I don't think this is a good solution. It probably has significant performance implications.

Instead, let's abstract the number formatting logic in `Size` into a utility (and let's not include any internationalization in the utility — maybe the utility can return english strings like `GB` and `MB` to denote the unit, and then we'll map these english units to the necessary translation objects in the component). Then, both the `Size` component and this new component can use the utility for getting the appropriate numerical values. Then the values can be passed through `this.props.intl.formatNumber`, and the units can be mapped to the translation object and then get passed to `this.props.intl.formatMessage`.

D3X: Your assumptions are absolutely correct. I didn't like this approach either, but decided to submit the PR for discussion anyway. Didn't want to refactor anything without your approval of the idea :)

I think what you've described makes sense. I'll try to implement it soon.

jfurrow:
@D3X Nicely done! Sorry for how long I've taken to review this.

Thank you for accepting and implementing my feedback. Double thanks for following the patterns I've established — even though some of them are dumb patterns that I'd like to change. Triple thanks for your patience!

I'm going to merge this as-is, but I do have some more feedback. If I had reviewed this PR in a timely manner, I'd ask you to make these changes, but instead I'll make the changes immediately after merging.

1. Flood ships with transpiled static assets (for now — this will be changing soonish), so JavaScript and CSS should be transpiled by running `npm run build-assets` any time the client source changes.
2. I don't like that `DocumentTitle` wraps the entire app — that means that every time state changes in `DocumentTitle`, the entire app also re-renders. I'd prefer to render `DocumentTitle` as a sibling of the routes, so that React doesn't have to diff the rest of the app when the title updates.

client/javascript/util/size.js
```diff
@@ -0,0 +1,51 @@
...
+  const UNIT_TO_STRING_ID = {
+    B: 'unit.size.byte',
+    kB: 'unit.size.kilobyte',
+    MB: 'unit.size.megabyte',
+    GB: 'unit.size.gigabyte',
+    TB: 'unit.size.gigabyte',
+  };
+
+  return UNIT_TO_STRING_ID[unit] || 'unit.size.byte';
+};
```
jfurrow: Great job refactoring this until a simple util!

D3X:
> Flood ships with transpiled static assets

Yes, I know. I intentionally didn't include those compiled files in the PR as it would conflict with any other updates. I think these should only be updated when things change in `master`. Of course, ideally they would not live in the git repo at all :)

Issue #282, opened 2017-03-03
Top toolbar / table header not visible on Chrome Android


rarescosma:
Android version: 6.0.1
Chrome version: 56.0.2924.87

The Chrome navigation bar obscures the flood top toolbar and/or table header. 
Please see the attached screenshot and let me know if you need any extra info to reproduce the issue. 
![screenshot_20170303-122918](https://cloud.githubusercontent.com/assets/520445/23547826/621cae00-000d-11e7-93fc-25e753f105ae.png)

jfurrow:
The entire UI is not mobile friendly at this point, but I do plan to make it so in the future.

na-ji:
How about some basic bootstrap magic ? Making the left side a collapsable menu and maybe also collapse the top buttons ?

berot3:
would love to see a responsive flood! Please make it happen!

I just discoverd that rutorrent has even 2 mobile-plugins! [1](https://github.com/xombiemp/rutorrentMobile) and [2](https://github.com/JimmyLaurent/ru-mobile/)

But with your design it would look much cooler ofc :smile:

Issue #348, opened 2017-05-08
[Bug] Safari on IOS missing topbar


xInsertx:
Running iOS 10.3

Don't see the top bar. Sometimes I see part of it.

![img_0268](https://cloud.githubusercontent.com/assets/10349764/25801446/f7b4ec6e-3430-11e7-81f4-5339885e2c17.PNG)
![img_0269](https://cloud.githubusercontent.com/assets/10349764/25801445/f7b2cd9e-3430-11e7-87e4-844ec7ee76a2.PNG)

jfurrow:
Yeah, admittedly I have only been testing Flood in Chrome. This will be trivial to fix, just need to get around to it...

geo5275:
This also have been happening for a while on Chrome for Android .

I'm running Chrome 58.0.3029.83 and Android 7.1.2

![screenshot_20170513-214012](https://cloud.githubusercontent.com/assets/20895147/26030295/c6727290-3825-11e7-8ed7-2fddc81b6505.png)
![screenshot_20170513-214040](https://cloud.githubusercontent.com/assets/20895147/26030294/c6726f20-3825-11e7-88e5-c858b9d7ae67.png)

disrupted:
has this been resolved? currently running into a similar issue with mobile Safari on iOS 12.1

PR #455, opened 2017-09-25
Fixed toolbar not being visible on mobile browsers.

D3X:
`vh` unit on mobile browsers includes the width of the address bar, so the toolbar (and part of the torrent list header) are rendered beneath it. On regular websites you can simply scroll and the bar disappears. Not so much with `flood` (and a lot of other similar apps, I guess). Changing the height of the right section from `100vh` to `100%` fixes the issue on mobile and does not change anything on desktop, because everything is absolutely positioned anyway.

Closes: #282

### DIFF
--- a/client/sass/base/_layout.scss
+++ b/client/sass/base/_layout.scss
@@ -27,7 +27,7 @@ body {
     display: flex;
     flex: 1 0 auto;
     flex-direction: column;
-    height: 100vh;
+    height: 100%;
     justify-content: center;
     position: relative;
   }


D3X:
I think it would also close #348 as it seems to be the same bug, but I don't have an iOS device to test it.

jfurrow:
Who takes over two weeks to review a single line of code? :wave:

jfurrow:
Thanks for this — seems to be working fine. The app needs a real mobile view at some point, but this is definitely helpful.

D3X:
> Who takes over two weeks to review a single line of code? 👋

Busy people? :)

PR #469, opened 2017-10-13
Ignore package-lock.json

varyoo:
I don't know what this is except it being created by `npm install`, it blocks my flood gitmodule from being staged by git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 /server/db
 /server/assets/*.map
 /server/temp
+package-lock.json


blakef:
Try `npm install --production --no-package-lock`, newer version of npm to [allow for reproducible dependency trees](https://docs.npmjs.com/files/package-lock.json).  This is something they'd probably want to commit to repo instead of adding it to the `.gitignore`.

jfurrow:
Yes, the package lock file should definitely be committed. Will be committed in https://github.com/jfurrow/flood/pull/417.

Issue #275, opened 2017-02-26
Crash when downloading via UI


bengadbois:
On commit `ca3fbbaece833ac42d3ade1b9b5e8315be3fb5b7`. Downloading 900MB files.

stdout says
```
GET /api/client/torrents/taxonomy 304 12.301 ms - -
buffer.js:11
    super(arg1, arg2, arg3);
    ^

RangeError: Array buffer allocation failed
    at Buffer.Uint8Array (native)
    at FastBuffer (buffer.js:11:5)
    at createUnsafeBuffer (buffer.js:38:12)
    at allocate (buffer.js:181:12)
    at Function.Buffer.allocUnsafe (buffer.js:141:10)
    at new Buffer (buffer.js:78:19)
    at ReadStream.<anonymous> (/home/x/flood/node_modules/archiver-utils/index.js:39:15)
    at emitNone (events.js:91:20)
    at ReadStream.emit (events.js:185:7)
    at endReadableNT (_stream_readable.js:974:12)

npm ERR! Linux 3.14.32-xxxx-grs-ipv6-64
npm ERR! argv "/home/x/.nvm/versions/node/v6.9.5/bin/node" "/home/x/.nvm/versions/node/v6.9.5/bin/npm" "start"
npm ERR! node v6.9.5
npm ERR! npm  v3.10.10
npm ERR! code ELIFECYCLE
npm ERR! flood@0.0.0 start: `node server/bin/www`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the flood@0.0.0 start script 'node server/bin/www'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the flood package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     node server/bin/www
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs flood
npm ERR! Or if that isn't available, you can get their info via:
npm ERR!     npm owner ls flood
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     /home/x/flood/npm-debug.log
```

npm-debug.log says
```
0 info it worked if it ends with ok
1 verbose cli [ '/home/x/.nvm/versions/node/v6.9.5/bin/node',
1 verbose cli   '/home/x/.nvm/versions/node/v6.9.5/bin/npm',
1 verbose cli   'start' ]
[...]

PR #474, opened 2017-10-17
pipe the download directly

dcousens:
Resolves https://github.com/jfurrow/flood/issues/275
And stops my Flood from dying (halting) for 20GiB+ files.

Increases throughput slightly too.

### DIFF
--- a/server/models/client.js
+++ b/server/models/client.js
@@ -101,24 +101,14 @@ var client = {
           }
         } else {
           const tempFilename = `${hash}-${Date.now()}.tar`;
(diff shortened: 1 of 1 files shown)


dcousens:
This also stops the "delay" that can be felt with larger files while node creates and copies the temp files...

jfurrow:
Oh wow, I didn't realize this would be so simple, but it definitely seems like the preferred solution. I'll do a bit of testing before merging — thanks for this!

jfurrow:
Thanks @dcousens this is a major improvement!

PR #475, opened 2017-10-17
darker menu item foreground

dcousens:
On different LCD screens,  I couldn't actually read this text, as it is too light.

The new colour is taken from the side bar.


jfurrow:
Definitely more legible. Thanks!

Issue #281, opened 2017-03-02
Option to add 'Download File' to torrent context menu


wopian:
I've recently ditched Deluge WebUI in favour of this on my server, which also required a seperate frontend to *access* the files to download - which Flood is capable of, so I've ditched that too.

However the only way I've found to download files is rather slow and tedious, as its behind 2 menus for each torrent.

1. Right click torrent (or double click - skip 2)
2. Click Torrent Details in context menu
3. Click Files tab
4. Check each file to download
5. Click download file link

Would be nice to have `Download File' in the right click context menu, as almost all of my torrents contain a single file. 

Plus it serves multiple file downloads in a tar archive, so you'd only have to use the File tab if you only want certain files.

xiprox:
A "copy link address" context menu item would be appreciated too.

jfurrow:
Thanks for the issue, I'll add this!

dcousens:
This would be amazing :+1:

PR #476, opened 2017-10-17
add download link

dcousens:
Resolves https://github.com/jfurrow/flood/issues/281

Implementation may leave something to desire... but works fantastic with https://github.com/jfurrow/flood/pull/474

### DIFF
--- a/client/src/javascript/components/torrent-list/TorrentList.js
+++ b/client/src/javascript/components/torrent-list/TorrentList.js
@@ -330,7 +330,7 @@ class TorrentListContainer extends React.Component {
     event.preventDefault();
     const baseURI = ConfigStore.getBaseURI();
     let link = document.createElement('a');
-    link.download = `${torrent.name}.tar`;
+    link.download = torrent.isMultiFile ? `${torrent.name}.tar` : torrent.name;
     link.href = `${baseURI}api/download?hash=${torrent.hash}`;
     link.click();
   }
--- a/client/src/javascript/components/torrent-list/TorrentList.js
+++ b/client/src/javascript/components/torrent-list/TorrentList.js
@@ -239,7 +239,7 @@ class TorrentListContainer extends React.Component {
       clickHandler,
       label: this.props.intl.formatMessage({
         id: 'torrents.list.context.move',
-        defaultMessage: 'Set Download Location'
+        defaultMessage: 'Set Torrent Location'
       })
     }, {
       type: 'separator'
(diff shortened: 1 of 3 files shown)


jfurrow:
Thanks @dcousens, I'll test this before merging. Something does feel a little dirty by creating the anchor element and calling its `click` method, but maybe it's totally fine.

dcousens:
The GC appears to clean it up as it is never added to the document heirarchy;  but my testing was only of about ~100mb worth of divs (took about 10 minutes!)... I don't think it will be an issue.

client/src/javascript/components/torrent-list/TorrentList.js
```diff
@@ -252,6 +253,15 @@ class TorrentListContainer extends React.Component {
         defaultMessage: 'Torrent Details'
       })
     }, {
+      action: 'torrent-download-tar',
+      clickHandler: (action, event) => {
+        clickHandler(action, event, torrent);
+      },
+      label: this.props.intl.formatMessage({
+        id: 'torrents.list.context.download',
+        defaultMessage: 'Download .tar'
```
jfurrow: What do you think about changing this label to `Download`? If it's a single file torrent, then it won't be archived.

client/src/javascript/components/torrent-list/TorrentList.js
```diff
@@ -313,6 +326,15 @@ class TorrentListContainer extends React.Component {
     });
   }
 
+  handleTorrentDownload(torrent, event) {
+    event.preventDefault();
+    const baseURI = ConfigStore.getBaseURI();
+    let link = document.createElement('a');
+    link.download = `${torrent.name}.tar`;
```
jfurrow: There's an `isMultiFile` property on the `torrent` object — would you mind checking for that here, and adjusting the `download` property accordingly?

E.g.
```javascript
link.download = torrent.isMultiFile ? `${torrent.name}.tar` : torrent.name;
```

client/src/javascript/components/torrent-list/TorrentList.js
dcousens: I don't mind.

`Download` is confusing anyway,  as we have `Set Download Location`...

Maybe if we changed that `Set Torrent Data Location` it would be OK?

jfurrow: Yes, `Set Torrent Data Location` is a lot more reasonable. Good idea!

client/src/javascript/components/torrent-list/TorrentList.js
dcousens: Fixed!

dcousens:
@jfurrow added b2dddf6 for the label change,  `Set Torrent Location` seemed cleaner... but, up to you.

Can change to `Set Torrent Data Location` if you think otherwise

jfurrow:
@dcousens I agree again, `Set Torrent Location` is cleaner. 🥇

Issue #473, opened 2017-10-17
package-lock.json is overwritten by npm install


varyoo:
So package-lock.json has been added to the repository, I hurry to test it:

~~~ txt
$ git clone https://github.com/jfurrow/flood
$ cd flood
$ npm install
$ git status
modified:   package-lock.json
~~~

I can observe in this repository 's package.json what is described in this answer on why `npm install` would overwrite package-lock: https://stackoverflow.com/a/45566871

midzer:
Perhaps some1 forgot to push latest `package-lock.json` to repo.

mirolm:
Is it okay config.js and alot of server side javascript to be included into main.js in assets?

jfurrow:
@varyoo It does make sense to heed the advice in the stackoverflow post, and use exact versions. I'll make this change.

@mirolm You're right that `config.js` should probably not be in the bundle, but none of the server side JavaScript is included.

jfurrow:
@varyoo I decided _not_ to use specific versions of dependencies, in favor of more up-to-date dependencies. I'll keep `package-lock.json` for now, and when the `--from-lock-file` option is supported, we'll recommend this in the readme.

PR #478, opened 2017-10-18
Use exact dependency versions

jfurrow:
Closes https://github.com/jfurrow/flood/issues/473


dcousens:
The issue is bothersome,  but I'd rather have up-to-date dependencies no?

jfurrow:
@dcousens that's a good point, and I didn't consider it. Is it better to have more reproducible builds, or up-to-date dependencies? I'm not really sure...

jfurrow:
Ok, I'm convinced we don't need the specific versions. I like the idea of using `--from-lock-file` when it's implemented. https://github.com/npm/npm/issues/18286

PR #487, opened 2017-10-21
[RSS Feeds] Always get items from enclosures

SanPilot:
The current behavior of the Feed service is to skip torrents from the enclosures if there is a `link` element. However, on many feeds, this doesn't work correctly.

For example, [the Arch Linux RSS feed](https://www.archlinux.org/feeds/releases/) uses the `link` tag to link to a web page with information about the download. The actual torrent URLs are in the enclosure tags.

This PR changes the behavior to include items from both the link tag and the enclosure tags.

### DIFF
--- a/server/services/feedService.js
+++ b/server/services/feedService.js
@@ -147,14 +147,18 @@ class FeedService {
 
   // TODO: Allow users to specify which key contains the URLs.
   getTorrentUrlsFromItem(feedItem) {
+
+    // Combine items from link tag and enclosures
+    let items = [];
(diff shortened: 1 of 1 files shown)


SanPilot:
Actually, this approach has a major flaw. Closing.

dcousens:
What was the flaw?

SanPilot:
Well, in most cases, this would match links that are to listing pages instead of the torrent file. I've opened a different PR that approaches this problem better.

PR #488, opened 2017-10-21
[RSS Feeds] Prefer enclosures over link tag

SanPilot:
Changes behavior of Feed Service to only use link tag if there are no enclosures.

This makes sense because trackers that use enclosures put the torrent's URLs in them (usually, the `link` tag is for humans). If the tracker doesn't have enclosures, the `link` tag is where the URL goes.

This is also how [ruTorrent's RSS plugin does it](https://github.com/Novik/ruTorrent/blob/c5bc4f7831b568084fff0cff8b52223ee61ee335/plugins/rss/rss.php#L209)

### DIFF
--- a/server/services/feedService.js
+++ b/server/services/feedService.js
@@ -147,10 +147,6 @@ class FeedService {
 
   // TODO: Allow users to specify which key contains the URLs.
   getTorrentUrlsFromItem(feedItem) {
-    if (feedItem.link) {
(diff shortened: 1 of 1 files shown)


jfurrow:
@SanPilot Thanks very much for the PR, and for providing the link to ruTorrent's logic. Flood's should be further refactored to better match ruTorrent's, but this is helpful.

Issue #491, opened 2017-10-21
Color toolbar buttons and remove pause.


mirolm:
Coloring toolbar buttons will make flood more intuitive. For most people red and green is associated with start and stop actions. 
- torrent start button - make it green on hover.
- torrent stop button - make it red on hover.
- remove torrent pause button - it does not work anyway. can be replaced with create torrent - mktorrent console utility can be called to do so.
- add torrent button - make it green on hover.
- delete torrent button - make it red on hover.

PR #494, opened 2017-10-21
Remove pause

dcousens:
Additionally bundled removing some unused dependencies as I went through the respective files.
If you want me to drop that commit, by all means not an issue.
I simply use [`standard`](https://www.npmjs.com/package/standard) locally, so it throws errors in my editor for these things.

Partially resolves https://github.com/jfurrow/flood/issues/491


dcousens:
Tested locally with no issues found.

jfurrow:
Thanks for the cleanup!

PR #495, opened 2017-10-22
use tar-stream directly to avoid OOM

dcousens:
Another follow up to https://github.com/jfurrow/flood/pull/474.

`archiver` would use [`collectStream`](https://github.com/archiverjs/archiver-utils/blob/master/index.js#L27-L49) to aggregate a file if it didn't know the file size before hand.

This would accumulate each file in memory before streaming it!
Bah!
Our approach was fine for many _medium_ files,  but failed for large files.

This PR avoids that by doing an `fs.stat` before creating a read stream,  and passing that `stat.size` directly to `tar-stream`.

I don't think this changes any assumptions,  in fact,  at least this way we get an **actual error** if the bytes written and bytes read mismatch.

### DIFF
--- a/server/models/client.js
+++ b/server/models/client.js
@@ -1,23 +1,20 @@
 'use strict';
 
-const archiver = require('archiver');
 const fs = require('fs');
 const path = require('path');
-const rimraf = require('rimraf');
-const util = require('util');
+const series = require('run-series');
+const tar = require('tar-stream');
 
 const ClientRequest = require('./ClientRequest');
 const clientResponseUtil = require('../util/clientResponseUtil');
 const clientSettingsMap = require('../../shared/constants/clientSettingsMap');
-const formatUtil = require('../../shared/util/formatUtil');
-const TemporaryStorage = require('./TemporaryStorage');
 const torrentFilePropsMap = require('../../shared/constants/torrentFilePropsMap');
 const torrentPeerPropsMap = require('../../shared/constants/torrentPeerPropsMap');
 const torrentService = require('../services/torrentService');
 const torrentTrackerPropsMap = require('../../shared/constants/torrentTrackerPropsMap');
 
 var client = {
-  addFiles: (req, callback) => {
+  addFiles (req, callback) {
     let files = req.files;
     let path = req.body.destination;
     let isBasePath = req.body.isBasePath === 'true';
@@ -53,7 +50,7 @@ var client = {
     });
   },
 
-  addUrls: (data, callback) => {
+  addUrls (data, callback) {
     let urls = data.urls;
     let path = data.destination;
     let isBasePath = data.isBasePath === 'true';
@@ -67,7 +64,7 @@ var client = {
     request.send();
   },
 
-  checkHash: (hashes, callback) => {
+  checkHash (hashes, callback) {
     let request = new ClientRequest();
 
     request.checkHash({hashes});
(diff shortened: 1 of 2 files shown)


server/models/client.js
```diff
@@ -1,23 +1,20 @@
...
-const rimraf = require('rimraf');
-const util = require('util');
+const series = require('run-series');
+const tar = require('tar-stream');
 
 const ClientRequest = require('./ClientRequest');
 const clientResponseUtil = require('../util/clientResponseUtil');
 const clientSettingsMap = require('../../shared/constants/clientSettingsMap');
-const formatUtil = require('../../shared/util/formatUtil');
-const TemporaryStorage = require('./TemporaryStorage');
```
dcousens: unused

server/models/client.js
```diff
@@ -1,23 +1,20 @@
...
 const torrentFilePropsMap = require('../../shared/constants/torrentFilePropsMap');
 const torrentPeerPropsMap = require('../../shared/constants/torrentPeerPropsMap');
 const torrentService = require('../services/torrentService');
 const torrentTrackerPropsMap = require('../../shared/constants/torrentTrackerPropsMap');
 
 var client = {
-  addFiles: (req, callback) => {
+  addFiles (req, callback) {
```
dcousens: formatting change inlined,  if no good I will remove

server/models/client.js
```diff
@@ -78,12 +75,14 @@ var client = {
...
   },
 
-  downloadFiles(hash, fileString, res) {
+  downloadFiles (hash, fileString, res) {
     try {
       const selectedTorrent = torrentService.getTorrent(hash);
       if (!selectedTorrent) return res.status(404).json({error: 'Torrent not found.'});
 
       this.getTorrentDetails(hash, (torrentDetails) => {
+        if (!torrentDetails) return res.status(404).json({error: 'Torrent details not found'});
```
dcousens: if the UI wasn't refreshed... explosions

server/models/client.js
```diff
@@ -100,37 +99,46 @@ var client = {
...
-          if (fs.existsSync(file)) {
-            res.attachment(path.basename(file));
-            res.download(file);
-          } else {
-            res.status(404).json({error: 'File not found.'});
-          }
-        } else {
-          const archive = archiver('tar', {store: true});
+          res.attachment(path.basename(file));
+          return res.download(file);
```
dcousens: refactored for early-exit

server/models/client.js
```diff
@@ -184,7 +192,7 @@ var client = {
         let value = datum[0];
         let settingsKey = clientSettingsMap[requestedSettingsKeys[index]];
 
-        if (!!outboundTransformation[settingsKey]) {
+        if (outboundTransformation[settingsKey]) {
```
dcousens: redundant negation

server/models/client.js
```diff
@@ -287,13 +295,9 @@ var client = {
...
 
-  setSettings: (payloads, callback) => {
+  setSettings (payloads, callback) {
     let request = new ClientRequest();
-
-    if (payloads.length === 0) {
-      callback({});
-      return;
-    }
+    if (payloads.length === 0) return callback({});
```
dcousens: is this meant to be `callback(null)` btw?

server/models/client.js
jfurrow: <3

server/models/client.js
jfurrow: No idea why I passed the empty object here. The UI doesn't do anything with this.

jfurrow:
@dcousens This is awesome!

PR #496, opened 2017-10-22
Implement flood-ui-kit

jfurrow:
This basically replaces all of the form elements with those in [flood-ui-kit](https://github.com/jfurrow/flood-ui-kit). Along the way I made some minor design adjustments and cleaned up the codebase a bit.

If anyone's trying to follow along, sorry for the giant PR!


jfurrow:
😬 Alright, let's do this.

dcousens:
The [slight] `background-color` changes for modals look a little bit... less polished?
But the tags being lighter helps in readability.

<3 the hover effects for list items.

dcousens:
tone
@jfurrow each Modal panel is now 1 color,  not 2 like in this

What is is now:

jfurrow:
@dcousens The single-tone modal is intentional. They could use a more significant redesign (e.g remove the need for these large modals entirely), but I think this is an improvement.

dcousens:
@jfurrow you could replace the throttles with a "custom" menu item in the throttle drop down,  which opens a pure-text modal with a "save" to enter the above data.

dcousens:
Is the slot information *ever* changed in a live manner?

PR #502, opened 2017-10-25
*Hopefully* fix enclosures

SanPilot:
I failed to test my last PR properly.

When I deployed new version of Flood to box, I realized that the enclosures issue remained. This PR should finally put that to rest.

### DIFF
--- a/server/services/feedService.js
+++ b/server/services/feedService.js
@@ -149,8 +149,8 @@ class FeedService {
   getTorrentUrlsFromItem(feedItem) {
     // If we've got an Array of enclosures, we'll iterate over the values and
     // look for the url key.
-    if (feedItem.enclosures && Array.isArray(feedItem.enclosures)) {
-      return feedItem.enclosures.reduce(
+    if (feedItem.enclosure && Array.isArray(feedItem.enclosure)) {
+      return feedItem.enclosure.reduce(
         (urls, enclosure) => {
           if (enclosure.url) {
             urls.push(enclosure.url);


jfurrow:
Sorry @SanPilot but I don't think this is going to work.

In the feed someone provided, the `enclosures` element was an array, and each item in the array was an element with attribute `url`, which contained the URL of the torrent.

In another feed I saw, there was no `enclosures` element, but rather a singular `enclosure` element, which had an attribute `url` containing the torrent's URL.

So, to make this function work with both the `enclosures` array and `enclosure` object, I think we should rewrite this function as follows:
```js
getTorrentUrlsFromItem(feedItem) {
  // If we've got an Array of enclosures, we'll iterate over the values and
  // look for the url key.
  if (feedItem.enclosures && Array.isArray(feedItem.enclosures)) {
    return feedItem.enclosures.reduce(
      (urls, enclosure) => {
        if (enclosure.url) {
          urls.push(enclosure.url);
        }

        return urls;
      },
      []
    );
  }

  // If there's a singular enclosure and it has a url key, let's use it.
  if (feedItem.enclosure && feedItem.enclosure.url) {
    return [feedItem.enclosure.url];
  }

  // If there are no enclosures, then use the link tag instead
  if (feedItem.link) {
    return [feedItem.link];
  }

  return [];
}
```

jfurrow:
@SanPilot I left a note about what we need to do to more closely match the ruTorrent plugin's behavior, here: https://github.com/jfurrow/flood/issues/461#issuecomment-339559068

For now, I think my suggestion above will get these feeds working for the feeds you specified. But maybe I'm wrong.

SanPilot:
Let me work on this some more

Issue #504, opened 2017-10-25
Incorrect filesize reporting on very large files


meoberto:
I have reached past the TB of upload in two torrents, however, as we can see here, the uploaded data size is displayed incorrectly, something which is evident in the Flood screenshot below, where I sorted my torrents by descending uploaded data.
![screenshot_20171025_101527](https://user-images.githubusercontent.com/30121325/31987927-a8fec6f4-b96d-11e7-84d1-93832270c25d.png)

On these top two torrents, rTorrent does however report the uploaded data correctly
![screenshot_20171025_101252](https://user-images.githubusercontent.com/30121325/31987954-b90c0d40-b96d-11e7-9514-c3b547dd6dd3.png)

Basically TB file sizes are being displayed as GB.

PR #505, opened 2017-10-25
Fixed sizes > 1TB displaying as GB.

D3X:
While refactoring the `Size` component into a util, I must've missed changing one translation string after copy-pasting it from somewhere else.
Here comes the fix.

Closes: #504


jfurrow:
Good catch @D3X and thanks for reviewing @dcousens!

PR #528, opened 2017-11-11
Use width instead of transform in ProgressBar

### LINKED ISSUE #517: Progress bar inconsistent height
Proof:
proof
OS:
Win 10 Pro Insider Preview, Build 17025
Browser: Chrome Canary Version 64.0.3259.0 64Bit

jfurrow:
Closes #517 (hopefully)

* Uses `width` instead of `transform` and removes the `transition`
* Removes progress bar caching, as it was only caching a few DOM nodes — not sure why I thought this would be helpful


dcousens:
Cannot reproduce #517 after this patch.

Issue #535, opened 2017-11-12
floodServerProxy & 502 Bad Gateway


noraj:
@jfurrow what's up with bd8a5fe152c06cf0d359521774678ab60251c18e, what is `floodServerProxy: 'http://127.0.0.1',` for?

I tried to removed it, put it to `floodServerProxy: 'http://127.0.0.1',` and `floodServerProxy: 'https://127.0.0.1',` but none worked, I always get `502 Bad Gateway` from nginx. Of course I restarted flood service each time I made a change. And also restarted nginx.

I currently don't know if it's a flood issue or a nginx one due to a nginx update. But as my others virtual servers are working and flood is the only one that is proxified and you introduce a new `floodServerProxy`, so I think the new parameter in `config.js` is in cause.

noraj1337:
I just found that there was a wiki page [wiki](https://github.com/jfurrow/flood/wiki/Using-Flood-behind-a-reverse-proxy) for that. My config is different but worked before.

Actually in `config.js` I have `baseURI: '/',` and my nginx config for flood looks like:

```
location / {
  proxy_pass https://127.0.0.1:3000;
  proxy_http_version 1.1;
  proxy_set_header Upgrade $http_upgrade;
  proxy_set_header Connection 'upgrade';
  proxy_set_header Host $host;
  proxy_cache_bypass $http_upgrade;
 }
```

So what do I need to set for `floodServerProxy`? Do I need the rewrite rule?

dcousens:
``` nginx
        upstream flood { server 127.0.0.1:9900; }

        # snip

        server {
                listen 80;
                server_name _;

                location / {
                        proxy_pass http://flood;
                        proxy_max_temp_file_size 0;
                }
        }
```

The re-write rule is unnecessary.

noraj1337:
Thanks @dcousens .So my nginx config should be good, and in fact I'm sure it is because it was working until a few days ago.

jfurrow:
>what's up with [bd8a5fe](https://github.com/jfurrow/flood/commit/bd8a5fe152c06cf0d359521774678ab60251c18e), what is floodServerProxy: 'http://127.0.0.1', for?

[...]

PR #540, opened 2017-11-16
Remove baseURI

dcousens:
RE https://github.com/jfurrow/flood/issues/535#issuecomment-344823126

`nginx.conf`
``` nginx
        upstream flood { server 127.0.0.1:3000; }

        # snip

        server {
                listen 80;
                server_name _

                location / {
                        return 200 "<h1>Foobar</h1>";
                        add_header Content-Type text/html;
                }

                location /flood/ { # slash here
                        proxy_pass http://flood/; # and slash here
                        proxy_max_temp_file_size 0; # for downloads > 1.1G
                }
        }
```

Tested and verified locally,  good luck @jfurrow :grinning:


client/src/javascript/stores/ConfigStore.js
```diff
@@ -26,7 +9,7 @@ class ConfigStoreClass extends BaseStore {
   }
 
   getBaseURI() {
-    return this.userConfig.baseURI || '/';
+    return '';
```
dcousens: Minimal change was the easiest

jfurrow:
@dcousens I'm super confused... Did you read my comment? baseURI is required for to properly isolate Flood to a subdirectory on a domain.

jfurrow:
Ohhh, my apologies — I see how this could work.

client/src/javascript/stores/ConfigStore.js
jfurrow: Yeah, this PR should be updated to remove all references to `baseURI`.

dcousens:
@jfurrow sorry for lackluster communication on my behalf,  I assumed it might be easier to demonstrate concretely than to make assumptions that I might be wrong about.
This PR simply shows that it works!  But,  open to discussion of course :+1:

dcousens:
> I see how this could work.

tl;dr - URLs are relative if there is no `/` on the front.

jfurrow:
> tl;dr - URLs are relative if there is no / on the front - so it will "just work".

This is only true if the routes aren't nested (they aren't right now, to be fair). As soon as we wish to create a nested route, this will break.

I don't think this is a good idea. It's not difficult for users to specify the `baseURI` option. I'll close this PR, but if you'd like to continue the discussion, you can open a ticket and we can gather more opinions there.

dcousens:
> This is only true if the routes aren't nested (they aren't right now, to be fair). As soon as we wish to create a nested route, this will break.

I don't think I follow,  do you mean something as in `/user/foo/profile`?
The router used on the client won't have any issues handling this as it is handled internally,  not by the server?

jfurrow:
> Any reason not to continue the discussion here?

It seems like this would be more discoverable by users in the future if it were taking place in an issue rather than a PR. But I'm okay with discussing here if that's what you'd prefer.

> The router used on the client won't have any issues handling this as it is handled in the router, not by the server?

The problem is with the requests' relative URLs and nested client-side routes. If the client-side's route is `/user/foo/profile`, and a request is initiated with relative URL `api/auth/logout`, then the actual request URL will be `/user/foo/profile/api/auth/logout`, so it fails. I tested this quickly with your PR by duplicating the `overview` route in `app.js` and changing it to `overview/foo/bar/baz` (of course I also had to modify the generated `index.html` file to use relative URLs). It failed to load, but maybe there's a solution for this — maybe the nginx config needs something else? I'm using the example you provided in this PR.

> I suppose my question is, why do you think the existing approach is a good idea?

So far, the only reason I think this approach is a good idea is because it circumvents the nested client-side route problem.

> I've never had to do this with other services

I'm very much in favor of following any best-practices or commonly implemented patterns for this. I'll do some research, but if you have any examples of documentation around this type of thing, I'd love to read them.

> it is really nice to be able to simply change my nginx configuration and not have to re-compile flood assets, nor need to edit the configuration.

That is understandable, but I honestly don't think it's that bad. Users are required to recompile static assets for every update already.

jfurrow:
One solution to this would be to use [`HashRouter`](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/api/HashRouter.md) instead of [`BrowserRouter`](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/api/BrowserRouter.md), but then we get an ugly `#` in the URL. I'd prefer not to do this.

dcousens:
Thanks for the thorough explanation @jfurrow :+1: ,  I'll check out the issues above and see if there exists a canonical approach :)

jfurrow:
I'm still convinced we need `baseURI`.

dcousens:
I haven't found a better solution yet either :+1:

PR #544, opened 2017-11-20
change default ssl behaviour

noraj1337:
using clear traffic (http) is a bad practice, default should be using ciphered traffic (https)


dcousens:
NACK,  I don't use `express` to do my SSL

noraj1337:
> NACK, I don't use express to do my SSL

_NACK_ do you mean [this](https://www.urbandictionary.com/define.php?term=nack) (I'm not native English speaker)?

And there is no correlation between between my PR (changing default behavior from http to https) and you talking about `express`. You may wanted to comment another issue maybe ?

dcousens:
Apologies,  NACK is "no-ACK", where ACK is acknowledge,  aka,  I don't agree with this PR.

I use `nginx` for my SSL termination,  and not the nodejs `https` module (via [`express`](https://github.com/jfurrow/flood/blob/master/server/app.js#L8)).

noraj1337:
@dcousens I also use `nginx` as revert proxy to expose flood on internet with a very strong ssl policy. But that doesn't mean that the local traffic between flood and nginx should be in clear. If there is a vulnerability on your server and an attacker gain an unprivileged user access, he will be able to see localhost unciphered connections. That's not the case if you use ssl.

jfurrow:
If we're to do this, don't we also need to supply (or generate) a certificate and key?

dcousens:
> But that doesn't mean that the local traffic between flood and nginx should be in clear. 

How is the Attacker able to intercept this without `root`?

@noraj1337 and even if the attacker can intercept that traffic,  and this encrypts that, but how do you prevent the attacker simply querying the `rtorrent` SCGI port?

noraj1337:
@jfurrow You are totally right, if we make this change we need to provide a default cert and key.

@dcousens I'm no saying it's perfect, I'm saying it's better than nothing. :/

dcousens:
>  I'm saying it's better than nothing. :/

And I'm saying,  I don't think it is.  It might provide a false sense of security where a user could be focusing their efforts elsewhere.
Perhaps isolating accounts and services instead.

noraj1337:
> Perhaps isolating accounts and services instead.

I gave an example of that on the wiki https://github.com/jfurrow/flood/wiki/Install-flood-with-a-service-account

dcousens:
@noraj1337 I understand,  and that documentation is fantastic.  I worry that this `SSL` might fool users otherwise into not bothering with that because "its encrypted".

SanPilot:
It can't hurt, at least, to use self-signed certificates for some encryption. I propose that Flood is set to use HTTPS by default and certificates are automatically generated with the build script (i.e. `npm run build`), if this is possible with node.

jfurrow:
I agree with @SanPilot — if we can reliably generate certificates when needed, then we can enable `https` by default. Until then, we can't.

In the interim, maybe we could just display a really big, obnoxious warning in the console.

@noraj1337 Would you prefer I keep this PR open?

noraj1337:
@jfurrow Creating self signed certificate in a shell is really easy, but I think this can't be done in pure JavaScript. Is creating a bash script and calling it as a sub-process in JavaScript an acceptable solution? We may close this issue if we open one like _Generating default SSL certificate_, what do you think?

dcousens:
@noraj1337 would `nginx` care about the self-signed certificate?

SanPilot:
See [pem](https://www.npmjs.com/package/pem).

SanPilot:
@dcousens By default, [no](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_ssl_verify).

dcousens:
I still don't understand the use case for this locally though.

You need `root` to listen to the packets... if they can `root`, we lose.

What are we protecting???

noraj1337:
@dcousens Yes you're right. The only case that matters is when users will serve flood directly with nodejs without using a reverse-proxy like nginx. It will be better form them to auth in https rather than in clear. Of course it's better to run it through nginx.

You make me think that we may need to create a _security_ section in the doc, telling to use a service account to run flood, using SSL, proxifying it, etc...

jfurrow:
>Is creating a bash script and calling it as a sub-process in JavaScript an acceptable solution?

If we're going to enable SSL by default, yes, this is acceptable to me (though [pem](https://www.npmjs.com/package/pem) sounds promising).

> We may close this issue if we open one like _Generating default SSL certificate_, what do you think?

Agreed — I'll close and open that issue.

> I still don't understand the use case for this locally though.

Agreed, and I don't think this is should be a high priority. But HTTPS by default would be great, I think, so long as it's really easy to set up.

> You make me think that we may need to create a _security_ section in the doc, telling to use a service account to run flood, using SSL, proxifying it, etc...

This is a really good idea.

Issue #537, opened 2017-11-15
Check hash is not working


ginja85:
For cross seeding I often use the force recheck in rutorrent, is such a feature available in flood and if not could it be added?

Thanks.

jfurrow:
@ginja85 Right-click on a torrent and click `Check Hash`

ginja85:
I've tried check hash, which doesn't seem to do anything, file location is correct, if I go to rutorrent and click force recheck that works.

noraj1337:
I can see this king of log:

```
Nov 27 18:56:33 hostname npm[10172]: POST /api/client/torrents/check-hash 200 13.565 ms - 13
```

But yes nothing happen or at least files are not checked.

PR #557, opened 2017-11-28
Fix hash checking experience

jfurrow:
## Description
1. When initiating a file hash check, Flood will tell rtorrent to "stop" the selected torrents (by invoking RPC commands `d.stop` and `d.close`). If the torrent was _not_ previously considered `stopped` (by [these disorderly conditions](https://github.com/jfurrow/flood/blob/dc0fea927967ea1197b855f290769a2a9047ab5d/server/services/torrentService.js#L153-L169), then it will "start" the selected torrents (by invoking RPC commands `d.open` and `d.start` immediately after `d.check_hash`).
2. Flood previously differentiated between "stopped" and "paused" torrents based on the `d.is_open` response, but with this PR, it will consider "paused" torrents to be "stopped" as well, which seems to better represent their behavior in rtorrent. I'm not sure if this is the best idea, or if the traits that make up the "paused" status could be meaningful — it's unclear to me what `d.is_open` represents. Probably it's something to do with the filesystem? [These docs](https://github.com/mdevaev/emonoda/wiki/rTorrent-XMLRPC-Reference) offer this vague description: `Get the state of the torrent (0=closed, 1=open)`. I'm open to suggestions here; surely `d.is_open` presents something meaningful, but from what I can tell, the Flood UI doesn't need to communicate this.

These two changes fix the bug with Flood's "check hash" logic for non-"stopped" torrents and the erroneous "downloading" state in the UI after performing "check hash" on a stopped torrent.

## Related Issue
Closes https://github.com/jfurrow/flood/issues/537

## Motivation and Context
The current behavior is broken as described in the issue.

## How Has This Been Tested?
Tested manually against "stopped" and "downloading" torrents.

## Types of changes

## Checklist:

### DIFF
--- a/server/models/ClientRequest.js
+++ b/server/models/ClientRequest.js
@@ -185,15 +185,24 @@ class ClientRequest {
 
   checkHash(options) {
     const hashes = this.getEnsuredArray(options.hashes);
-    const hashesToStop = hashes.filter(hash => {
+    const stoppedHashes = hashes.filter(hash => {
       return torrentService.getTorrent(hash).status.includes(torrentStatusMap.stopped);
     });
+    const hashesToStart = [];
+
+    this.stopTorrents({ hashes });
 
     hashes.forEach(hash => {
       this.requests.push(this.getMethodCall('d.check_hash', [hash]));
+
+      if (!stoppedHashes.includes(hash)) {
+        hashesToStart.push(hash);
+      }
     });
 
-    this.stopTorrents({hashes: hashesToStop});
+    if (hashesToStart.length) {
+      this.startTorrents({ hashes: hashesToStart });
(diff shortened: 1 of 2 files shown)


noraj1337:
1. It need to start it back only if the torrent was started when the `check` was launched. If torrent stopped and you want to check, you do not want the torrent to be started after that.

jfurrow:
@noraj1337 That's covered by this PR, as far as I know, so I think this is ready to merge!

noraj1337:
I tested that, that seems to work perfectly. But from the UI this is nice in Extended view but as in condensed view the percentage is not written so you don't know if it's 100% or 99,8% and need to check in detailed view.

Issue #390, opened 2017-06-25
Torrents with multiple files aren't handled correctly


JohnFlowerful:
When adding a torrent with multiple files (e.g. [this](http://linuxtracker.org/index.php?page=torrent-details&id=a35898f236132f3fad41b445473dc6a206700827)), the parent directory for the torrent is not created. Instead the torrent's files are created in the directory specified by 'Destination'.

The torrent directory can be created with 'Set Download Location', but that leads me to another issue: When using 'Set Download Location' with move data ticked, the files are not moved. It appears they are copied. It also appears that this function starts the torrent even if the previous state was stopped.

Setup:
libtorrent-ps feature-bind ([ebuild](https://github.com/JohnFlowerful/flowerlay/blob/master/net-libs/libtorrent/libtorrent-9999.ebuild))
rtorrent-ps feature-bind ([ebuild](https://github.com/JohnFlowerful/flowerlay/blob/master/net-p2p/rtorrent/rtorrent-9999.ebuild))
nginx running for reverse proxy: https://p.teknik.io/Raw/RJu6R
Flood config: https://p.teknik.io/Raw/IuIVk

rtorrent displays no errors, neither does npm nor client (Chromium 61.0.3133.0).

P.S. sorry for the multi-issue issue~

varyoo:
I observed the same issue since the "Base path" option has been added.

jfurrow:
Sorry for the delay in this silly bug!

hunti88:
Hello, with commit 524a0796c9bcf95693f4f3fe0ae1ad8633441d47, this is broken again.

jfurrow:
Sorry @hunti88, will take a look.

varyoo:
Dupe https://github.com/jfurrow/flood/issues/506

noraj1337:
I confirm that parent directory is not created (making a mess in the root torrent folder), the `Move data` doesn't works, and even if you move files manually the `Check Hash` option is not working (see issue #537). Managing Torrents with multiple files is very painful because of those issues and this can lead into ratio drop on private trackers if a mistake make you download files twice. So I think this is a critical bug.

dcousens:
Also confirmed here

PR #565, opened 2017-11-29
Fix erroneous stringified boolean value interpretation

jfurrow:
## Description
When uploading files, a [`FormData`](https://developer.mozilla.org/en-US/docs/Web/API/FormData) instance is used to submit the torrent files and options. `FormData.append` stringifies boolean values, so Flood interpreted `isBasePath: 'false'` as truthy, and used the destination as the torrent's base path.

* Adds `coerceBoolean` middleware
* Implements the middleware for `isBasePath` value

## Related Issue
Fixes https://github.com/jfurrow/flood/issues/390

## Motivation and Context
Bad bug

## How Has This Been Tested?
Added torrents by file & URL

## Types of changes

## Checklist:

### DIFF
--- a/server/middleware/booleanCoerce.js
+++ b/server/middleware/booleanCoerce.js
@@ -2,7 +2,7 @@
 
 module.exports = key => {
   return (req, res, next) => {
-    const value = req.body[key];
+    const value = req.body && req.body[key];
 
     if (value && typeof value === 'string') {
       req.body[key] = value === 'true';
--- a/server/middleware/booleanCoerce.js
+++ b/server/middleware/booleanCoerce.js
@@ -1,0 +1,13 @@
+'use strict';
+
+module.exports = key => {
+  return (req, res, next) => {
+    const value = req.body[key];
+
+    if (value && typeof value === 'string') {
+      req.body[key] = value === 'true';
+    }
+
+    next();
+  };
+};
--- a/server/routes/client.js
+++ b/server/routes/client.js
@@ -3,6 +3,7 @@ const express = require('express');
 const multer = require('multer');
 
 const ajaxUtil = require('../util/ajaxUtil');
+const booleanCoerce = require('../middleware/booleanCoerce');
 const client = require('../models/client');
 const clientRequestService = require('../services/clientRequestService');
 const router = express.Router();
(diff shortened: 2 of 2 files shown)


jfurrow:
@dcousens @noraj1337 for review, maybe?

server/middleware/booleanCoerce.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+module.exports = key => {
+  return (req, res, next) => {
+    const value = req.body[key];
```
dcousens: boom if `req.body` is undefined

server/routes/client.js
```diff
@@ -17,9 +18,14 @@ router.post('/add', function(req, res, next) {
   client.addUrls(req.body, ajaxUtil.getResponseFn(res));
 });
 
-router.post('/add-files', upload.array('torrents'), function(req, res, next) {
-  client.addFiles(req, ajaxUtil.getResponseFn(res));
-});
+router.post(
+  '/add-files',
+  upload.array('torrents'),
+  booleanCoerce('isBasePath'),
```
dcousens: do you really need to make this a middleware?

Why not
``` js
if (req.body) {
  req.body.isBasePath = req.body.isBasePath === 'true'
}
```

server/middleware/booleanCoerce.js
jfurrow: `req.body` will always be defined because of the `body-parser` middleware.

server/routes/client.js
jfurrow: That's very specific — I like the middleware approach because it's easy to use elsewhere.

server/middleware/booleanCoerce.js
jfurrow: I added a check for this anyway.

server/routes/client.js
dcousens: I wouldn't think this is a common problem.
Why is it happening in the first place?

jfurrow: This is not common, that's true! 

This is happening because a [`FormData`](https://developer.mozilla.org/en-US/docs/Web/API/FormData) instance is used to upload the binary torrent files, and that `FormData` instance also includes the options (one of the options is this `isBasePath`), and `FormData.append` stringifies boolean values.

dcousens: In that event,  I would be specific,  this is clearly an edge-case?

jfurrow: It would happen any time `FormData` is used, so I think this should be flexible.

What do you think is the harm in creating and using this middleware?

dcousens: More places to look when debugging!
Do we need more `FormData` instances?

It is a bike-shed at this point,  if you prefer it,  leave it.

jfurrow: Fair point, and no, we don't need more `FormData` instances yet.

I do prefer it, so I'll just merge as-is, but I do appreciate your feedback!

PR #576, opened 2017-12-03
Updated Dockerfile

0xcaff:
## Description

This does two things:

1. No longer user onbuild. Onbuild is deprecated for [reasons outlined
   here][reasons].

2. Build static assets with the container. Before, the static assets wouldn't be built causing the container to crash on start up.

[reasons]: https://github.com/docker-library/official-images/issues/2076

## Related Issue
#48, #79, #304 

## Motivation and Context
The existing build wasn't working and was crashing upon container start.

## How Has This Been Tested?
I tested this locally by running:

    docker build -t flood .
    docker run --rm -it flood # with environment configuration variables

## Types of changes

## Checklist:


Dockerfile
```diff
@@ -1,5 +1,32 @@
-FROM node:7.8.0-onbuild
-RUN apt-get update && apt-get install -y mediainfo && rm -rf /var/lib/apt/lists/*
-COPY config.docker.js config.js
+FROM node:7.8.0
+
+# Install runtime dependencies.
+RUN apt-get update && apt-get install -y mediainfo
+
+# Remove state generated by apt-get update.
+RUN rm -rf /var/lib/apt/lists/*
```
mdarse: You should should merge this command with the previous one if you don't want the state to be still in an intermediary filesystem layer (or use [multi-stage builds](https://docs.docker.com/engine/userguide/eng-image/multistage-build/)).

Dockerfile
```diff
@@ -1,5 +1,32 @@
...
+
+# Copy docker configuration file. This allows for specifying common configs with
+# just environment variables.
+COPY config.docker.js /usr/src/app/config.js
+
+# Install Dependencies
+RUN npm install
+
+# Remove cache (to keep image small).
+RUN npm cache clean --force
```
mdarse: Same here

Dockerfile
0xcaff: Fixed. Didn't know about multi-stage builds!

Dockerfile
0xcaff: Fixed

jfurrow:
Thanks @0xcaff and @mdarse!

0xcaff:
Awesome project @jfuerow! Could you please enable automated builds on Docker Hub also?

Issue #582, opened 2017-12-10
Cannot Install on arm32v7/node:9.2.1-onbuild


toebbel:
Installation of argon2 dependency fails on arm architecture.

## Expected Behavior
- `npm install` for flood should work regardless of underlying platform

## Current Behavior

```
> argon2@0.16.2 install /usr/src/app/node_modules/argon2                                                                                      
> node-gyp rebuild                 

make: Entering directory '/usr/src/app/node_modules/argon2/build'      
  CC(target) Release/obj.target/libargon2/argon2/src/ref.o             
*** Error in `cc': double free or corruption (top): 0x01f62c38 ***     
libargon2.target.mk:106: recipe for target 'Release/obj.target/libargon2/argon2/src/ref.o' failed                                             
make: *** [Release/obj.target/libargon2/argon2/src/ref.o] Aborted (core dumped)                       
```

## Possible Solution
- drop argon2 dependency

## Steps to Reproduce (for bugs)
- `docker run -it arm32v7/node:9.2.1-onbuild bash` on a raspberry pi
- git clone this repo
- `npm install`

## Context
This is obviously not your fault. The maintainer of `argon2` do not support installation on all arm platforms (they made a recent change, which does not help in my situation). I would rather suggest to drop `argon2` in favor of `bcryptjs`, which is purely implemented in javascript and does not pull in any dependencies (and thus can not fail on arbitrary platforms). I can see that you migrated from `bcrypt` to `argon2` and you will have your reasons. I believe that you want to allow the broadest possible userbase to run flood. To achieve this goal, I suggest to flatten dependencies and drop native bindings.

## Your Environment

* Version used:  bba07ed  
* Environment name and version: node 9.2.1 on arm32v7
* Operating System and version: `Linux raspberrypi 4.9.41-v7+`

d0tfile:
I have raspberry pi 2 B+ with Arch Linux installed (4.9.67-1-ARCH, armv7l). I have no errors while running "npm install" or "npm run build". Are you sure it is argon2 issue? May docker cause this issue?

toebbel:
Which gcc compiler do you have on your raspberry? My build failed in the
node 9 Docker image. I can check later, which gcc compiler is in there.

dcousens:
@toebbel you should file this in the Argon2 repository... not here.

toebbel:
Hi @dcousens,
I agree that discussion makes no sense here after #583 was closed.
[...]

Issue #514, opened 2017-10-30
Modernize hashing password algorithm


noraj1337:
Issue https://github.com/jfurrow/flood/issues/483 :  _bcrypt via node-gyp fails to compile on FreeBSD_ was closed because randomnonsense found a way to compile on FreeBSD but there is still the discussion about changing from bcrypt to a more modern and more secure hashing password algorithm like Argon2.

Here were some of my comments on from isse #483 : 

https://github.com/jfurrow/flood/issues/483#issuecomment-338382112

> **TL;DR**: please use **Argon2** with [libsodium for nodejs](https://github.com/paixaop/node-sodium)
> 
> Blowfish is a deprecated cipher, openssh removed support for it in [last release](https://www.openssh.com/txt/release-7.6) because there are some well known collision attacks.
> Bcrypt is more robust than simple blowfish and is still acceptable for password hashing but it begins to be quite old and is based on blowfish which is outdated.
> The last [Password Hashing Competition winner](https://password-hashing.net/) is **Argon2**, this is the best secure way to store your password.
> 
> Please stop using bcrypt because blowfish support will be removed everywhere in the future. Don't use SHA-512 raw it's no designed for password storage an is [optimized for GPU](https://security.stackexchange.com/questions/133239/what-is-the-specific-reason-to-prefer-bcrypt-or-pbkdf2-over-sha256-crypt-in-pass).
> 
> SHA2 (sha-256, sha-512),  SHA3 (black2, keccak) are optimized to be fast but Bcrypt, Scrypt, PBKDF2, Argon2 are designed to be slow in order to slow attacker for password hashing/storage!
> 
> It's not up to date for algorithm to use be here are some good practices : [Salted Password Hashing - Doing it Right](https://crackstation.net/hashing-security.htm)
> 
> please refer to this : [How to Safely Store Your Users' Passwords in 2016](https://paragonie.com/blog/2016/02/how-safely-store-password-in-2016), here [anchor for nodejs](https://paragonie.com/blog/2016/02/how-safely-store-password-in-2016#nodejs).
> 
> I'm not a crypto expert just a cybersecurity engineer but my knowledge tell me that using the last winner of PHC is the best you can do.
> 
> PS : may be open a new issue to switch form bcrypt to argon2 

https://github.com/jfurrow/flood/issues/483#issuecomment-338460908

> **TL;DR** : PHD said Argon2, just use it.
> 
[...]

Issue #483, opened 2017-10-18
bcrypt via node-gyp fails to compile on FreeBSD


randomnonsense:
As explained here: https://github.com/jfurrow/flood/commit/d38ab8525275771e7ef52667e16403a36b836938

Which means most recent commits break flood, which was so far working perfectly fine. Thanks.

randomnonsense:
Can a workaround for this be dropping encryption completely? And make it user select-able when installing via 'npm install' for a custom non-blowfish deployment? I run everything in a secure environment anyway. I do "understand the risks".

jfurrow:
@randomnonsense I'm sorry that the commit is breaking for you. I switched to [this alternative bcrypt module](https://github.com/kelektiv/node.bcrypt.js) because it's actively maintained and well-trusted. I think you can agree this is good for the majority of users, especially those — like myself — who are not security experts.

Can you please look at these similar issues found on the Bcrypt module's repo? https://github.com/kelektiv/node.bcrypt.js/issues?utf8=%E2%9C%93&q=is%3Aissue%20freebsd In all of those issues, the users were able to solve the problem. There are no active issues reporting builds failing on FreeBSD, to my knowledge.

If that doesn't work, please file an issue in that repository, and in the meantime, consider using
@wonderfall's [fantastic Dockerfile](https://github.com/Wonderfall/dockerfiles/blob/master/rtorrent-flood/Dockerfile) (it should replace the one in the Flood repo, honestly).

randomnonsense:
> @randomnonsense I'm sorry that the commit is breaking for you. In the commit you referenced, I switched to this bcrypt module because it's actively maintained and well-trusted. I think you can agree this is good for the majority of users, especially those — like myself — who are not security experts.

Sorry, but I can't agree at all. It also doesn't seem you actually reviewed the problem. This feature was removed from FreeBSD's ABI on purpose as it was clearly deprecated. Consider https://reviews.llvm.org/D21329 which explains what's been done. So the change you made is not a change making a step forward, but backward. How is that "good for the majority of users" if Flood now uses a library based on deprecated premises?

Sadly this means no FreeBSD 11+ user will be able to use Flood. And that's that. FreeBSD is not a trivial system with irrelevant userbase.

Can you think of any way to work around this? Did you perhaps consider replacing blowfish with AES? Now, that would be an advancement.

jfurrow:
[...]

PR #583, opened 2017-12-10
use zero-dependency password hashing npm package

toebbel:
## Description
Replace `argon2` with `bcryptjs` to avoid native dependencies.

## Related Issue
- https://github.com/jfurrow/flood/issues/582

## Motivation and Context
This fixes installations on platforms where `argon2` does not build

## How Has This Been Tested?
1. create user
2. login with credentials
3. login with wrong credentials

## Types of changes
- [x] Breaking change (fix or feature that would cause existing functionality to change)

I did not add logic to migrate hashes from `argon2` to `bcryptjs`. Logins would be invalidated !!!

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,7 @@
     "abbrev": {
       "version": "1.1.1",
       "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
-      "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q=="
+      "integrity": "sha1-+PLIh60Qv2f2NPAFtph/7TF5qsg="
     },
     "accepts": {
       "version": "1.3.4",
@@ -137,7 +137,7 @@
         "string-width": {
           "version": "2.1.1",
           "resolved": "https://registry.npmjs.org/string-width/-/string-width-2.1.1.tgz",
-          "integrity": "sha512-nOqH59deCq9SRHlxq1Aw85Jnt4w6KvLKqWVik6oA9ZklXLNIOlqg4F2yrT1MVaTjAqvVwdfeZ7w7aCvJD7ugkw==",
+          "integrity": "sha1-q5Pyeo3BPSjKyBXEYhQ6bZASrp4=",
           "dev": true,
           "requires": {
             "is-fullwidth-code-point": "2.0.0",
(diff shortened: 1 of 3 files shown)


noraj1337:
We specially wanted to move from bcrypt to argon2, see those issues:
- [Modernize hashing password algorithm](https://github.com/jfurrow/flood/issues/514)
- [bcrypt via node-gyp fails to compile on FreeBSD](https://github.com/jfurrow/flood/issues/483) (ps: bcrypt also doesn't build on some platforms).

Sorry but I don't think we will merge this PR.

toebbel:
Thanks noraj1337 for your response. I can fully understand your motivation to keep dependencies updated and pick them based on security and performance among many criteria. I also understand, that some people in the community prefer argon2 over bcrypt2.

If I understand correctly, #483 is a similar issue as #582 (different hashing algorithms and different platforms, but same issue). However, this PR pulls in `bcryptjs*, which does not need to be compiled at all and will therefore work on *all* architectures that node runs on.

As the Javascript ecosystem moves relatively fast compared to others, I want to point out this blog post as reference, why it might be not ideal to jump on new encryption algorithms too soon.

> I also place bcrypt over Argon2, because Argon2 was just recently announced as the Password Hashing Contest winner. As with all cryptographic primitives, we need to time to analyze, attack, and pick apart the design. If after about 5 years, it still stands strong and secure, then it can be recommended as a solution for production. In the meantime, it's something certainly worth testing, but maybe not for production code.

https://pthree.org/2016/06/28/lets-talk-password-hashing/

I am sure, whatever choice you make, it's the way to go for flood.

noraj1337:
Your quote is the personal point of view of Aaron Toponce, that seems to be a Linux dev but not a cryptographer. And as I said in issue #514 : 
> Argon2 was the winner of PHC in July 2015 but was already finished in March 2014. PHC winner mean that the algorithm was chosen and tested by all the best crypto expert of the world.

Argon is enough reliable and bcrypt is deprecated. If you **fully** read issue #514 you will be able to convince yourself that the change was a need. Moving back to bcrypt is no way.

But if the problem is only compilation on some platform it is also possible to change from the argon npm package to one of the libsodium package to use argon2 algorithm that will not need compilation either. 

> This fixes installations on platforms where argon2 does not build

As far as I know, we do not currently have unsolved issue for argon2 compilation if was always due to to a bug in a specific node version or a user mistake. So affirming this seems not very useful.


> If I understand correctly, #483 is a similar issue as #582 (different hashing algorithms and different platforms, but same issue).

At the time of issue #483, PR #523 was not there yet but the [bcrypt](https://www.npmjs.com/package/bcrypt) package was already using `node-gyp`. I understand you want to use `bcryptjs` instead but as I said returning back to bcrypt is not acceptable. The only satisfying way seems to find a solution to use argon2 without compilation, and the 2 libsodium package I mentioned in my first comment seems to do that.

Sorry again if your work won't be merged. It's better to open a discussion issue before doing a breaking pull request in order to avoid rejected PR. I wish you didn't take to much time to make this PR.

All the discussion, reference, arguments, and decision about bcrypt vs argon2 are in the issue #514 and we won't come back on that.

dcousens:
Great discussion by all involved :+1:

PR #553, opened 2017-11-28
Make username case-insensitive

SanPilot:
See issue #552.

This PR makes usernames case-insensitive.

Checklist

- [x] Accept username with any case variations
- [x] Prevent creation of new users with existing usernames of any case

(Closes #552)

PR #586, opened 2017-12-17
Escape regex strings on authentication

### LINKED ISSUE #585: Usernames are interpreted as regexes on login
Since support for case insensitive usernames has been added, they are interpreted as arbitrary regexes due to the following code:

```javascript
this.db.findOne({username: {$regex: new RegExp(credentials.username, 'i')}}).exec((err, user) => {
```
You can login without knowing a valid username just by entering .* as username. Luckily, you still need to provide the correct password so it is not a total authentication bypass.

## Expected Behavior

You should only be able to login to an account by entering the full username.

## Current Behavior

You can enter .* or any other regex.

## Possible Solution

In my opinion, the change that added case-insensitive usernames should be reverted. I don't see any clear benefits by having them and it complicates user handling and thereby might introduce additional bugs further down the line.

## Steps to Reproduce (for bugs)

1. enter .* as username
2. enter the correct password
3. login successfully

SanPilot:
My previous PR #553 used the raw user input in the regex, making injection very easy.
## Description

Escapes characters including '.*\$' before querying the db.

## Related Issue

Closes #585 
## Motivation and Context

This fixes broken login behavior.
## How Has This Been Tested?

Tested using a local installation of Flood.

## Types of changes

## Checklist:

### DIFF
--- a/server/models/Users.js
+++ b/server/models/Users.js
@@ -11,8 +11,13 @@ class Users {
     this.db = this.loadDatabase();
   }
 
+  caselessUserRegex(user) {
+    var escapedUser = user.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, '\\$&'); // escape regex characters
+    return new RegExp('^' + escapedUser + '$', 'i');
+  }
+
   comparePassword(credentials, callback) {
-    this.db.findOne({username: {$regex: new RegExp(credentials.username, 'i')}}).exec((err, user) => {
+    this.db.findOne({username: {$regex: this.caselessUserRegex(credentials.username)}}).exec((err, user) => {
       if (err) {
         return callback(null, err);
       }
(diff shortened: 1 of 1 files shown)


wopian:
This is pretty slow for something that could be achieved with a single prototype method - it just has the downside of requiring the user database to be migrated on first run.

https://jsperf.com/regex-vs-tolowercase-2

server/models/Users.js
```diff
@@ -11,8 +11,13 @@ class Users {
     this.db = this.loadDatabase();
   }
 
+  caselessUserRegex(user) {
+    var escapedUser = user.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, '\\$&'); // escape regex characters
+    return new RegExp('^' + escapedUser + '$', 'i');
```
dcousens: This is a rabbit hole.

Stop searching using `$regex`.

jfurrow:
Thanks @SanPilot but I think I'd prefer to avoid the regex solution altogether.

Issue #628, opened 2018-02-25
[Enhancement] Stale bot for issue management


noraj1337:
In order to improve issue management, we need a stale bot.

## Expected Behavior

Abandoned issues should be closed.

## Current Behavior

Some issues are inactive for more than 1 year.

## Possible Solution

Install the [stale bot](https://github.com/probot/stale):  https://github.com/apps/stale

## Configuration

He is a proposition of configuration for the bot:

```yml
# Configuration for probot-stale - https://github.com/probot/stale

# Number of days of inactivity before an Issue or Pull Request becomes stale
daysUntilStale: 60

# Number of days of inactivity before a stale Issue or Pull Request is closed.
# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
daysUntilClose: 30

# Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
exemptLabels:
  - critical
  - security
  - bug

# Set to true to ignore issues in a project (defaults to false)
exemptProjects: false

# Set to true to ignore issues in a milestone (defaults to false)
exemptMilestones: false

# Label to use when marking as stale
staleLabel: stale

# Comment to post when marking as stale. Set to `false` to disable
markComment: >
  This issue has been automatically marked as stale because it has not had
  recent activity. It will be closed if no further activity occurs. Thank you
  for your contributions.

# Comment to post when removing the stale label.
# unmarkComment: >
#   Your comment here.

# Comment to post when closing a stale Issue or Pull Request.
# closeComment: >
#   Your comment here.

# Limit the number of actions per hour, from 1-30. Default is 30
limitPerRun: 30

# Limit to only `issues` or `pulls`
only: issues

# Optionally, specify configuration settings that are specific to just 'issues' or 'pulls':
# pulls:
#   daysUntilStale: 30
#   markComment: >
#     This pull request has been automatically marked as stale because it has not had
#     recent activity. It will be closed if no further activity occurs. Thank you
#     for your contributions.

# issues:
#   exemptLabels:
#     - confirmed
```

bbashy:
This is stale :)

noraj:
That's why it's needed :D but @jfurrow is not super hot with that feature.

bbashy:
I think, if you keep on top of issues, it should be fine.

I was only replying to try clear some of the issues out. I'll be trying to help others and contribute as I've switched from ruTorrent to Flood.

PR #629, opened 2018-02-25
Create stale.yml

noraj:
## Description

add `.github/stale.yml` for the. stale bot

## Related Issue

see https://github.com/jfurrow/flood/issues/628

## Motivation and Context

see https://github.com/jfurrow/flood/issues/628

## How Has This Been Tested?

N.A.

## Types of changes

## Checklist:

- [X] I have read the **CONTRIBUTING** document.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,55 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 60
+
+# Number of days of inactivity before a stale Issue or Pull Request is closed.
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 30
+
+# Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
+exemptLabels:
+  - critical
+  - security
+  - bug
+
+# Set to true to ignore issues in a project (defaults to false)
+exemptProjects: false
+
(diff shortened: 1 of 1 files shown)


noraj1337:
as said in see https://github.com/jfurrow/flood/issues/628, the bot need to be installed from here first https://github.com/apps/stale

jfurrow:
I really don't see the value here, can you try to convince me @noraj1337?

Even though issues get old and stale, the preservation of the discussion is is useful. Flood development is moving really slowly (unfortunately), which would mean a lot of these issues would be closed without being resolved.

noraj1337:
Here is how it works : `daysUntilStale` : time before being flagged as stale and then `daysUntilClose` time until it's closed since it entered into stale state.

As dev is slow we can increase the values, for example `daysUntilStale:120` (4 months) and `daysUntilClose:45`.

The goal is not to close the issue, in fact when an issue will be inactive for more than 4 months the bot will flag it as stale and put a message, so all people who are following this issue will get notified by github. These people will be able to say if they have more information about the issue, if they have a patch, an idea etc... or if the issue was solved by another one or is not happening anymore etc ... Then people can remove the stale flag or close the issue. There is only when nobody will care even after having been notified that the bot will close the issue (that's rare, for example when someone requesting a feature is not using the software anymore and that nobody else care).

So this is not for closing issues, but to get sure the issue is not forgotten, and to relaunch the debate.

So normally this will not hurt the preservation of issues but dynamism galvanize them.

@jfurrow What do you think? Are you convinced? :smile:

jfurrow:
I'm gonna pass on this for now.

Issue #164, opened 2016-11-01
Torrent are not added


Angristan:
Sometimes, when adding torrent it just does not work : 
![screenshot_01-11-2016_b gy7z9qpo](https://cloud.githubusercontent.com/assets/11699655/19910385/de2f5456-a08c-11e6-9ced-0d832d85f15e.png)

If I add multiple torrents, I will stay on the "Add Torrents" window, but the other torrents will be added.

Here is what nodejs tells us : 
```
Trace: { [Error: XML-RPC fault: Call XML not a proper XML-RPC call.  Call is not valid XML.  Invalid XML "content".  Problem with token at '<...': unclosed token]
  code: -503,
  faultCode: -503,
  faultString: 'Call XML not a proper XML-RPC call.  Call is not valid XML.  Invalid XML "content".  Problem with token at \'<...\': unclosed token' }
    at ClientRequest.onCompleteFn (/srv/seedbox/flood/server/util/ajaxUtil.js:7:17)
    at ClientRequest.handleError (/srv/seedbox/flood/server/models/ClientRequest.js:69:12)
    at _rejected (/srv/seedbox/flood/node_modules/q/q.js:844:24)
    at /srv/seedbox/flood/node_modules/q/q.js:870:30
    at Promise.when (/srv/seedbox/flood/node_modules/q/q.js:1122:31)
    at Promise.promise.promiseDispatch (/srv/seedbox/flood/node_modules/q/q.js:788:41)
    at /srv/seedbox/flood/node_modules/q/q.js:604:44
    at runSingle (/srv/seedbox/flood/node_modules/q/q.js:137:13)
    at flush (/srv/seedbox/flood/node_modules/q/q.js:125:13)
    at nextTickCallbackWith0Args (node.js:420:9)
POST /api/client/add-files 500 30.530 ms - 178

```

jfurrow:
Thanks for the issue, I've not seen this before.

Are you adding torrents by URLs or by uploading the torrent files?

Angristan:
In this case it's files, but I think it already happened with magnet, I'm not sure anymore

Flyy-y:
Hey, i'm getting the same issue, only with some torrents, adding with files.
![flood2](https://cloud.githubusercontent.com/assets/7773641/20032465/b7e672e2-a38a-11e6-8498-80fbdf8646b9.png)
[...]

PR #635, opened 2018-03-09
Fix wrong SCGI XML-RPC body length calculation method

crux153:
Fix wrong XML-RPC body length calculation method which results in SCGI calls failing when multibyte characters are included.

## Description
I discovered that flood calculates XML-RPC body length of an SCGI call simply by using String.prototype.length. This works ok when all payload is consist of single-byte characters. 

However, when any multibyte character is included in xml payload, it miscalculate body length as JavaScript's string is UTF-16 and every UTF-8 multibyte characters are single-byte in UTF-16. So while String.prototype.length reports that the length of a UTF-8 character is 1, it may be 2-4 bytes when encoded in UTF-8, which causes rTorrent's XML-RPC server to misbehave.

`[Error: XML-RPC fault: Call XML not a proper XML-RPC call.  Call is not valid XML.  Invalid XML "content".  Problem with token at '<...': unclosed token]`

This error is thrown when UTF-8 multibyte character is included in XML body, as current implementation of SCGI call builder reports shorter body length than its actual length.

I fixed this issue by calculating body length properly using Node's Buffer.byteLength function, which returns real byte length of UTF-8 encoded string.

## Related Issue
#164 #196 

I think changes in #196 may be reverted, but not tested.

## Motivation and Context
I faced this issue when I tried to add torrents with download path including CJK characters. I wasn't able to use flood since my hard drive is full of directories containing UTF-8 multibyte characters and flood couldn't handle it.

This PR fixes issues with CJK character handling.

## How Has This Been Tested?
I'm running the modified source on my machine for several days with bunch of torrents with multibyte character path. No problem so far.

## Types of changes

## Checklist:

### DIFF
--- a/server/util/scgi.js
+++ b/server/util/scgi.js
@@ -33,11 +33,12 @@ const sendMethodCall = (methodName, parameters) => {
     const deserializer = new Deserializer('utf8');
     const stream = net.connect(connectMethod);
     const xml = Serializer.serializeMethodCall(methodName, parameters);
+    const xmlLength = Buffer.byteLength(xml, 'utf8');
 
     stream.setEncoding('UTF8');
 
     const headerItems = [
-      `CONTENT_LENGTH${nullChar}${xml.length}${nullChar}`,
+      `CONTENT_LENGTH${nullChar}${xmlLength}${nullChar}`,
       `SCGI${nullChar}1${nullChar}`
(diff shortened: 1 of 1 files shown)


jfurrow:
Good catch @crux153, thanks for the PR!

PR #549, opened 2017-11-25
Multi rtorrent instances support

zawapete:
## Description
- Added rtorrent scgi data on user signup
- Lazy load user/instance db with multi session support
- Add username to xml-rpc callback
- Users data separation in services

## Related Issue
https://github.com/jfurrow/flood/issues/460#issuecomment-341980797

## Motivation and Context
See issue

## How Has This Been Tested?
Tested on a server with three separated rtorrent instances on socket

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

Once it works we could provide a migration script so that the change won't be breaking

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

Issue #460, opened 2017-10-09
Separate download folders for users


sachaw:
The ability have each users torrents separate is needed.

jfurrow:
Yeah, this is a good idea. I'm still not sure how best to handle multiple users...

sachaw:
I think it could be as simple as adding an user id to each torrent, then setting the basedir to /downloads/`username` or something like that, so when the user logs in it will display torrents that have there userid

varyoo:
What's wrong with a Flood instance per user?

dcousens:
@varyoo that would be a `rtorrent` *and* `flood` instance per user?

varyoo:
Yes of course!

sachaw:
Very heavy on resources, terrible idea.

varyoo:
>Very heavy on resources, terrible idea.

Please be more specific sir.

wopian:
Requiring a separate instance of rtorrent & flood for each user does not scale at all and wastes excessive amounts of memory (rtorrent uses 40mb and flood 60mb when *idle*).

Thus the lowest tier from DigitalOcean would only be able to physically allow 5 users (with 0 memory left to do anything). The tier can easily support 20-30 users with a single instance and wouldn't ever have to worry about memory usage

varyoo:
@wopian how many torrents for those values?

wopian:
0

sachaw:
Yea, separate instances are not the the way to go about this, like I said earlier, assigning the id of the user who added the torrent to the torrent in a db then calling all torrents with the logged in users id and display them, then we can have a single instance of rtorrent and flood.

jfurrow:
A long time ago I think I suggested that multi-users be handled by multiple instances of rtorrent. I agree this is a bad idea.

I think true multiuser support would be cool, but would be a huge amount of work, and it's not a big priority for me. If someone else wants to take it on, then let's come to an agreement about how it should be implemented first.

jvacek:
(summary of #518) What about multiple rtorrent instances and one flood server that can have multiple users that can each just connect to the right socket?
This will allow separate settings, torrents, home folders, limits, labels, etc.

jvacek:
The #549 PR is good, has anyone else tested it out yet? nudge nudge :)

PR #638, opened 2018-03-14
Multi rtorrent instances support

jfurrow:
## Description
This is the feature branch for multi-user support. Thanks to @zawapete for the hard work, which landed in https://github.com/jfurrow/flood/pull/549.

**What's left to do**
- [x] UI polish
- [x] Migration script
- [ ] Thoroughly review and clean up code
- [ ] Update documentation

I'll open PRs against this feature branch until I think it's ready.

## Related Issue
https://github.com/jfurrow/flood/issues/460#issuecomment-341980797

## Motivation and Context
See issue

## How Has This Been Tested?
- [x] Locally by @zawapete (they wrote: `Tested on a server with three separated rtorrent instances on socket`)
- [ ] Locally by @jfurrow

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have updated the documentation accordingly.


server/models/ClientRequest.js
```diff
@@ -183,26 +184,32 @@ class ClientRequest {
...
-    const stoppedHashes = hashes.filter(hash => {
-      return torrentService.getTorrent(hash).status.includes(torrentStatusMap.stopped);
-    });
-    const hashesToStart = [];
+  checkHash(userId, options) {
+    const torrentService = ServicesHandler.getTorrentService(userId);
+    torrentService.fetchTorrentList().then(
+      () => {
+        const hashes = this.getEnsuredArray(options.hashes);
+        const stoppedHashes = hashes.filter(hash => {
```
jfurrow: There's no need to create this array; we can just check `torrentService.getTorrent(hash)` in the `forEach`.

server/models/ClientRequest.js
```diff
@@ -183,26 +184,32 @@ class ClientRequest {
...
+          if (!stoppedHashes.includes(hash)) {
+            hashesToStart.push(hash);
+          }
+        });
+
+        if (hashesToStart.length) {
+          this.startTorrents({hashes: hashesToStart});
+        }
+      }
+    ).catch((err) => console.log(err));
```
jfurrow: Let's kill the `console.log`. Also we're not actually handling the error — this method would have thrown previously.

server/models/HistoryEra.js
```diff
@@ -45,6 +47,18 @@ class HistoryEra {
     this.startAutoCleanup(cleanupInterval, this.db);
   }
 
+  loadDatabase(userId, dbName) {
+    let dbPath = `${config.dbPath}${userId}/`;
+
+    let db = new Datastore({
```
jfurrow: This should be `const`, but let's just return this after setting `this.ready`.

server/models/HistoryEra.js
```diff
@@ -45,6 +47,18 @@ class HistoryEra {
     this.startAutoCleanup(cleanupInterval, this.db);
   }
 
+  loadDatabase(userId, dbName) {
+    let dbPath = `${config.dbPath}${userId}/`;
```
jfurrow: This should be `const`, but let's just concatenate the string all at once in the `filename` key assignment.

server/models/Users.js
```diff
@@ -96,12 +106,25 @@ class Users {
...
+      if (user == null) {
+        return callback(null, user);
+      }
+
+      this.db.remove({username: username}, {}, (err, numRemoved) => {
+        if (err) {
+          return callback(null, err);
+        }
+
+        fs.removeSync(`${config.dbPath}${user._id}/`);
```
jfurrow: This should be asynchronous.

server/models/client.js
```diff
@@ -12,15 +12,15 @@ const settings = require('./settings');
...
 var client = {
-  addFiles (req, callback) {
+  addFiles (userId, req, callback) {
     let files = req.files;
     let path = req.body.destination;
     let isBasePath = req.body.isBasePath;
-    let request = new ClientRequest();
+    let request = new ClientRequest(userId);
     let start = req.body.start;
     let tags = req.body.tags;
```
jfurrow: Need to introduce lint rules to address all of this nonsense...

server/models/client.js
```diff
@@ -31,114 +31,125 @@ var client = {
...
       fileRequest.addFiles({files: file, path, isBasePath, start, tags});
 
       // Set the callback for only the last request.
       if (index === files.length - 1) {
         fileRequest.onComplete((response, error) => {
-          torrentService.fetchTorrentList();
-          callback(response, error);
+          torrentService.fetchTorrentList().then(
+            () => callback(response, error)
+          ).catch((err) => console.log(err));
```
jfurrow: Let's kill the `catch` unless we're actually going to handle this error.

server/models/ClientRequest.js
```diff
@@ -183,26 +184,32 @@ class ClientRequest {
     });
   }
 
-  checkHash(options) {
-    const hashes = this.getEnsuredArray(options.hashes);
-    const stoppedHashes = hashes.filter(hash => {
-      return torrentService.getTorrent(hash).status.includes(torrentStatusMap.stopped);
-    });
-    const hashesToStart = [];
+  checkHash(userId, options) {
```
jfurrow: Each instance of `ClientRequest` receives the `userId` when it was instantiated, and it is stored on the instance as property `userId`. So none of the instance methods need to receive `userId` as an argument, they can just access the property at `this.userId`.

server/models/HistoryEra.js
```diff
@@ -45,6 +47,18 @@ class HistoryEra {
     this.startAutoCleanup(cleanupInterval, this.db);
   }
 
+  loadDatabase(userId, dbName) {
```
jfurrow: As I commented on `ClientRequest`'s `checkHash` method , let's retrieve `userId` on the instance rather than this extra argument.

server/models/Users.js
```diff
@@ -77,10 +85,12 @@ class Users {
       return callback(null, 'Username cannot be empty.');
     }
 
+    const socket = 0 < socketPath.trim().length;
```
jfurrow: This will throw when `socketPath` is `null` or `undefined`

server/models/Users.js
```diff
@@ -96,12 +106,25 @@ class Users {
   }
 
   removeUser(username, callback) {
-    this.db.remove({username: username}, {}, (err, numRemoved) => {
+    this.db.findOne({username: username}).exec((err, user) => {
```
jfurrow: Do we need to call `findOne` first?

server/models/ClientRequest.js
zawapete: `torrentService` could have been loaded but haven't yet fetched the torrent list (since it doesn't call `fetchTorrentList()` in the constructor)

server/models/ClientRequest.js
zawapete: you're right

server/models/HistoryEra.js
zawapete: agree

server/models/HistoryEra.js
zawapete: it was basically the same method as it was before, but for the `userId` :)

server/models/Users.js
zawapete: this would be true for every `const` extracted from `credentials`

server/models/Users.js
zawapete: we need the `user._id` in order to remove the db directory

server/models/Users.js
zawapete: totally asynchronous or waiting anyway for the method callback?
also, probably we should delete the directory first and then remove the user from the db. so if any error occurs we end up in a probably recoverable status

jfurrow:
Just to clarify, @zawapete, I left these comments mostly for myself — please don't feel obligated to implement anything here! I've started making these changes locally and will finish this up shortly.

zawapete:
@jfurrow , sure, I got that, I only left some comments to clarify why I wrote some code in the way I did :)

PR #676, opened 2018-06-02
Multi-user cleanup

jfurrow:
## Description
This PR cleans up the multi-user branch a bit.

* Services are instantiated at startup for all users (one service per user, even if multiple users share the same rtorrent instance)
* Service instances are added to Express' request object at `req.services`
* `clientRequestManager` has been introduced to throttle requests to the client on a per-user basis (should really be per-rtorrent-instance, but that should be part of a separate and thoughtful refactor)
* Service instances were modified to store the user object when instantiated and consume it in all of their methods, which restores a lot of these methods to their equivalents on `master`
* `path` module is used to construct filesystem paths for databases 😬 
* Removes some unnecessary `use strict` directives (will remove the rest separately)

## Motivation and Context
* Ensures multiple rtorrent instance connections are compatible with Flood's existing architecture
* Reduces complexity of controlling multiple rtorrent instances

## How Has This Been Tested?
Locally 🤷‍♀️

## Types of changes

## Checklist:


zawapete:
good for me

Issue #720, opened 2018-09-27
[Discussion] Stuff unrelated to a feature mixed in a PR


noraj:
## Discussion

@jfurrow On your PR **Multi rtorrent instances support** (#638), you made changes to the README unrelated to this PR that should have the object of a separate PR.

I just saw this today 26 days after it was merged.

For example this note was removed:

> **Note**: each time you modify `baseURI` in `config.js` you need to recompile assets (`npm run build`). To be sure follow the *Updating* procedure each time you modify the `config.js` file.

I don't know if it is still pertinent.

But more importantly those two notes where removed:

> * Note: Since #523 [node-gyp](https://www.npmjs.com/package/node-gyp) is needed. **IF** you need `sudo`, use `sudo npm i --unsafe-perm` (see [here](https://github.com/nodejs/node-gyp/issues/454) for why `--unsafe-perm`) else installation will fail. If you dont need sudo just use `npm i` as usual.
> * If your system use python3 as default you will need to install python2 and use `npm i --python="/usr/bin/python2"`.

Maybe the sudo stuff shouldn't be mentioned but we should say the when running `npm install` one **MUST** not put `sudo` before and make sure he has permission to rerad/write on that folder by using chown/chmod before.

Also the python stuff is maybe still up to date.

You also removed

> Build the documentation `npm run build-docs`.

Killing the idea of creating a documentation.

You also removed some docker env var but didn't reflected the change to the wiki https://github.com/jfurrow/flood/wiki/Docker

Then you also removed screenshots link without creating a Screenshot wiki page, loosing the screenshots.

Also those changes are highly unrelated to the Multi rtorrent instances support.

Please keep the changes in a PR related to the feature, don't mix stuff. Else it is hard to review and to find from where did a change happened. Also when you make a feature or change some documentation make sure to reflect it in the whole project.

jfurrow:
I don't see this as a problem at all.

The architecture of the application was significantly changed in [that PR](https://github.com/jfurrow/flood/pull/638), so the readme needed a thorough revamp. I removed some outdated information and bad troubleshooting advice for atypical environments, but I think that's completely warranted — the commit sat in review, for a long time, with a reasonable name of ["Updates documentation"](https://github.com/jfurrow/flood/pull/638/commits/6c363bb37931110fa07def424bbb46a562049aeb).

I agree with the sentiment that the scope of PRs should be as small as possible, but I'm also okay with some cleanup along the way, especially in an isolated commit.

[...]

PR #724, opened 2018-09-30
add back a removed step

noraj:
## Description
Add back documentation building step.

## Related Issue

#720


noraj:
@jfurrow This is just a line added back on the README, there is not test needed. You can review and merge that in a second.

jfurrow:
I don't think this belongs in the readme. If anything the docs should be generated in CI.

noraj:
@jfurrow I don't remember where we talk about that but yes I agree with you doc should be automatically build with the CI and serve to a github page.

Like I did for https://gitlab.com/noraj/nvd_api, automatically building the doc with [gitlab-ci](https://gitlab.com/noraj/nvd_api/blob/master/.gitlab-ci.yml) and serving it with gitlab pages.

But I thought this was a good point to add a step on how to build it locally for developers that did changes to the documentation and want to check how it looks.

Issue #654, opened 2018-04-16
Add a source/about link


noraj:
## Summary

The Flood UI lack of a pointer to its source repository.

## Expected Behavior

There should be an *about*/*author*/*github* page or button.

## Current Behavior

There is no *about*/*author*/*github* page or button.

## Possible Solution

One of the two solutions below:

1) Add a *github* icon next to the *notification* icon with a link to https://github.com/jfurrow/flood

![](https://i.imgur.com/yrYggK7.png)

2) Add an *about* tab in the *settings* page.

![](https://i.imgur.com/hXwB076.png)

## Context

This is useful when you spotted a bug and want to quickly open an issue.

This is also useful for people who didn't installed Flood but get it from an out-of-box, so they can know where the UI comes from.

noraj:
I'm not a great fan of using a third party redirecter.
It's possible for people to drop referer from their reverse proxy.
Example for nginx: `proxy_set_header Referer "";`.

bbashy:
@noraj So you think just link directly?

noraj:
Yes, what do you think ?

bbashy:
Will update it. I guess people will copy the link anyway if they're worried about it.

noraj:
They will may be already using a plugin to drop referer in their browser anyway.

bbashy:
Yeah. Also forgot about `rel="noreferrer"` :) I already had that in there.

PR #727, opened 2018-10-03
add about tab to settings modal

bbashy:
## Description
Added an about page to the settings modal with some info taken from the README (now in an ABOUT.md).

## Related Issue
#654 

## How Has This Been Tested?
I've developed this with a fully running instance.

Will update with one.

## Types of changes

## Checklist:


bbashy:
Something has broken since I last did this (rebased recent changes) and I can't work out what

```
Error: TextRenderer(...): A valid React element (or null) must be returned. You may have returned undefined, an array or some other invalid object
```

client/src/javascript/components/modals/settings-modal/AboutTab.js
```diff
@@ -0,0 +1,23 @@
...
+
+import SettingsTab from './SettingsTab';
+
+export default class AboutTab extends SettingsTab {
+  state = {
+    about: null,
+  };
+
+  componentWillMount() {
+    fetch(AboutMarkdownPath)
```
jfurrow: 1. Is this going to work for users who have the `baseURI` config set in `config.template.js`? I can't remember if Webpack will handle this automatically not.
2. `fetch` is lovely but we should use `axios` for consistency — [here](https://github.com/jfurrow/flood/blob/master/client/src/javascript/actions/SettingsActions.js#L51-L53) is an example of a GET request.

bbashy: 1. This path should always be correct unless the path of `client/src/javascript/components/modals/settings-modal/AboutTab.js` moves.
2. Can't use axois for non-HTTP requests?

jfurrow: Webpack is treating this as a static external resource, it doesn't become part of the JS bundle. `fetch` is fetching the static resource and parsing it, so this should use `axios`.

bbashy: Makes sense now. It's actually calling it from a file.

I'm not sure if `baseURI` will affect this.

jfurrow:
@bbashy Changes look good but failing CI build

noraj:
Node 8 and 10 CI image seems broken:
image
image

Node 9 fails at build:
image

bbashy:
Can you retry them for this PR? They did this a few times - it's Travis, not the code.

jeanralphaviles:
Looks like this PR caused #784 due to the ABOUT.md reference in [AboutTab.js](https://github.com/Flood-UI/flood/blob/5065ab2f89ca3d306aaf3a299eedd3853859f1c0/client/src/javascript/components/modals/settings-modal/AboutTab.js#L5). I suspect that ABOUT.md needs to be copied over in the Dockerfile.

PR #622, opened 2018-02-16
Smaller Docker image

sgsunder:
## Description
1. This basically shrinks the final Docker image from ~900MB to ~300MB by using the Alpine Linux version of the official node image, and excluding the copy of some unnecessary files into the final image.
2. The node_modules folder is built in a separate image as part of a multi-stage build. 
  - This will take advantage of Docker's build cache so that node_modules isn't recreated every rebuild (unless package.json has been changed, of course). This is a problem on the current Dockerfile and leads to incredibly slow build times.
  - This allows for build dependencies (like python, gcc, etc.) to be skipped from the final image.

## How Has This Been Tested?
Tested on my seedbox machine and seems to run identically to the current image

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #738, opened 2018-11-06
Smaller docker image build.

alekc:
Better use of multistaging build.

## Description
Current dockerfile doesn't fully use advantages of multistaging (well, it actually has only 1 stage). 

Operation `npm prune --production` deletes files, but they are being left on layer. 

These are images weights. 
Current version:
```
REPOSITORY                     TAG                 IMAGE ID            CREATED             SIZE
flood                          latest              1b6a9b21add0        3 minutes ago       338MB
```

My version:
```
REPOSITORY                     TAG                 IMAGE ID            CREATED             SIZE
flood                          latest              bc958b0e88bd        2 minutes ago       248MB
```

## Related Issue
https://github.com/jfurrow/flood/pull/622

## Motivation and Context
Smaller filesize

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,10 @@
 ARG NODE_IMAGE=node:10.1-alpine
+ARG WORKDIR=/usr/src/app/
 
 FROM ${NODE_IMAGE} as nodebuild
+ARG WORKDIR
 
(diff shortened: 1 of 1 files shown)


noraj:
Waiting for @jfurrow review.

PR #743, opened 2018-11-24
fix: handle broken symlinks on TorrentDestination

blizzardo:
Fixes the bug that caused the `Filesystem.getDirectoryList` to break on non-existing symlinks.

## Description
When choosing a directory for torrent destination, if that directory contains a broken symlink, it results in the directory selector to display: `This path does not exist. It will be created`

This fix adds a `existsSync` check in `getDirectoryList` to verify that the file exists before pushing it to the array. 

## Related Issue
`Filesystem.getDirectoryList` breaks on files that do not exist. The `statSync` function is used to check if the item is a file or directory. If the item that is being check does not exist, such as a broken symlink, it results in the following error:

```Error: ENOENT: no such file or directory, stat '/home/anon/.steampath'```


noraj:
waiting for @jfurrow to review that.

Issue #405, opened 2017-07-18
High CPU usage


Soullivaneuh:
I have a rtorrent instance with currently ~2800 active torrents.

The node server is regularly at nearly 100% CPU usage:

![image](https://user-images.githubusercontent.com/1698357/28326764-f533a258-6be1-11e7-8597-1c84ec286091.png)

And this even if I don't have any opened browser tab on it.

Any idea of why?

jfurrow:
This is unfortunate, I suspect it's struggling to process 2800 torrents each time Flood gets the updated torrent list from rTorrent. What CPU are you running this on?

You can adjust how frequently Flood polls rTorrent by editing `torrentClientPollInterval` in `config.js`. This defaults to `2000` milliseconds (2 seconds). You can try to change this to 5 or 10 seconds, which would reduce the continuous load on your CPU, but won't solve the core problem.

To understand the core problem, let's look at what Flood does each time it requests data from rTorrent:
1. Ask rTorrent for list of torrents with all of the details needed by Flood ([here are all of the details requested](https://github.com/jfurrow/flood/blob/master/server/constants/torrentListPropMap.js)).
2. Transform rTorrent's response into an object of torrents, [here](https://github.com/jfurrow/flood/blob/master/server/services/clientRequestService.js#L171-L239). rTorrent's response is a two-dimensional array, each nested array being an array of values. This means each time Flood gets the list from rTorrent, it iterates over each torrent (2800 torrents in your case) and also iterates over each requested value, transforming the values as necessary ([torrentListPropMap](https://github.com/jfurrow/flood/blob/master/server/constants/torrentListPropMap.js) contains the functions which transform the values — 36 values in total, right now).
3. Some values need to be determined by one or more of the 36 values returned by rTorrent (percent complete and ETA, for example). So we loop over any external reducers ([here](https://github.com/jfurrow/flood/blob/master/server/services/clientRequestService.js#L206-L213)]. Currently, all external reducers are added [here](https://github.com/jfurrow/flood/blob/master/server/services/torrentService.js#L30-L43).
4. After this object of torrent data is created, Flood iterates over each torrent _again_, comparing the new set of torrent data with the old, constructing a diff with which the client is updated.

That's a lot of work for the server to be doing — at the very least 218,400 iterations every 2 seconds, if my math is correct (2800 torrents * 36 values + 2800 torrents * 3 external reducers + 2800 torrents * 39 total values for diff comparison). 

To optimize, we can do a number of things...
[...]

PR #748, opened 2019-01-01
Stop polling when there are no active connnections for user

duvholt:
## Description

Disable polling in torrent service and history service when a user is not connected. 
Added a "connection manager" to keep track of the number of connections each user has.
Currently I'm directly calling the fetch methods, but maybe it's smarter to add some more generic methods for starting and stopping polling that all the services could theoretically use. 

## Related Issue
#405

## Motivation and Context

The motivation behind this change is to reduce the idle CPU usage of flood. Currently flood will regularly poll rtorrent even if there are no active clients connected. 

## How Has This Been Tested?

This has only been tested locally with three users and one active rtorrent server.

## Types of changes

This might be a breaking change if something in flood depends on the always on background polling, but I'm not sure. If so I think it would be smart to add a new config option to turn on or off idle polling. 

## Checklist:

### DIFF
--- a/server/middleware/clientActivityStream.js
+++ b/server/middleware/clientActivityStream.js
@@ -16,6 +16,12 @@ module.exports = (req, res) => {
 
   const serviceInstances = services.getAllServices(user);
   const serverEvent = new ServerEvent(res);
+  if (!serviceInstances.connectionManager.hasConnections()) {
+    serviceInstances.torrentService.fetchTorrentList();
+    serviceInstances.historyService.fetchCurrentTransferSummary();
+  }
+  serviceInstances.connectionManager.connect();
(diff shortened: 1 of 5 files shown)


duvholt:
Installation of `geoip-country-only` fails on Node v10, but that shouldn't be related to my PR.

mirolm:
It works when you install it manually with "npm install geoip-country-only". It is the same with node 10 and 11...

wopian:
`geoip-country-only` is causing the build to fail since its trying to request data from a URL that was previously deprecated and now completely removed (as of 2nd Jan)

https://github.com/jfurrow/flood/issues/751#issuecomment-451021662

Issue #558, opened 2017-11-28
Incorrectly shows 100%


Avamander:
Flood incorrectly displays "Downloaded | 100%" but rtorrent reports otherwise (99,98% complete of a 100GB torrent). It would be nice if it didn't do any rounding in the precentage section in advanced details, having more decimal places wouldn't hurt anyone but would avoid all situations like these.

noraj1337:
Yes, and when no in the advanced details it's better to truncate `99,98%` into `99%` than round it up to `100%`.

Syc0x00:
@jfurrow I believe this issue is complete.

PR #764, opened 2019-02-08
Fix downloaded rounding showing 100% when not all pieces are downloaded

jmarinaro:
## Description

Truncates rather than rounds download percent on torrent details and file list

## Related Issue

#558

## Motivation and Context

Flood can show 100% downloaded when not all pieces are downloaded

## How Has This Been Tested?

Ran flood in docker, created torrents with files of various sizes using `dd if=/dev/zero`

old details
old details
old filelist
old filelist
new details
new details
new filelist
new filelist

## Types of changes

## Checklist:

### DIFF
--- a/server/util/truncateUtil.js
+++ b/server/util/truncateUtil.js
@@ -1,0 +1,6 @@
+const truncateUtil = function(num, fixed = null) {
+  var re = new RegExp('^-?\\d+(?:\\.\\d{0,' + (fixed || -1) + '})?');
+  return num.toString().match(re)[0];
+};
+
+module.exports = truncateUtil;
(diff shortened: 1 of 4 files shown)


noraj:
@jmarinaro Thanks; we are waiting for @jfurrow to review that.

server/util/truncateUtil.js
```diff
@@ -0,0 +1,6 @@
+const truncateUtil = function(num, fixed = null) {
```
jfurrow: Let's rename this file `numberUtils` and export a named function called `truncateTo`. And instead of using this regepx, I think we do something simple like:
```js
const truncateTo = (num, precision = 0) => {
  const factor = 10 ** precision;
  return Math.floor(num * factor) / factor;
};
```

I think this will do the same thing and be a little faster.

PR #769, opened 2019-03-06
add complete tutorial

noraj:


jfurrow:
I don't feel comfortable endorsing in the README, but you can add to the wiki if you want.

noraj:
@jfurrow I was may be a little to much euphoric, the wiki feels to be the right place.

PR #806, opened 2019-06-01
SCGI: Fix Promise Rejection

damolp:
## Description
* Fixes bug where if you specify an incorrect server connection it loads a blank flood page
* Fixes one of the unhandled promise rejection errors (There's still one more that I can't narrow down)
* `net.connect` seems like it might throw other errors so this may be better suited in a try / catch block? or a node.js domain?

## Related Issue
N/A. Let me know if you want me to open one.

### DIFF
--- a/server/util/scgiUtil.js
+++ b/server/util/scgiUtil.js
@@ -23,7 +23,8 @@ const methodCall = (connectionMethod, methodName, parameters) =>
     const stream = net.connect(networkConfiguration);
     const xml = Serializer.serializeMethodCall(methodName, parameters);
     const xmlLength = Buffer.byteLength(xml, 'utf8');
-
+  
+    stream.on('error', reject);
     stream.setEncoding('UTF8');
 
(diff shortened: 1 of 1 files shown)


server/util/scgiUtil.js
```diff
@@ -34,7 +35,7 @@ const methodCall = (connectionMethod, methodName, parameters) =>
 
     bufferStream(stream).then(data => {
       rTorrentDeserializer.deserialize(data, resolve, reject);
-    });
+    }).catch(err => reject(err));
```
jfurrow: Can be simplified to `catch(reject)`

jfurrow:
> Fixes one of the unhandled promise rejection errors (There's still one more that I can't narrow down)

I'll try to help debug

> `net.connect` seems like it might throw other errors so this may be better suited in a try / catch block? or a node.js domain?

Looks like `domain` is for Node clusters, which Flood doesn't use, so a `try`/`catch` seems more appropriate if something here is expected to throw an error synchronously

jfurrow:
I'm actually going to accept as-is and I'll clean up afterwards, thanks @damolp!

Issue #308, opened 2017-04-18
Disk usage


RC1140:
Hi
I think it would be quite beneficial to have a small section on the sidebar with the servers disk stats (Free/Used basically), have started working on this for myself but let me know if there is a specific direction or feature set you think should be included.

Thanks for the awesome project.

wopian:
👍, would let me remove the remaining disk space column and only have that information shown once

jfurrow:
Good idea, and thanks for your willingness to implement this! I haven't put a ton of thought into this, but here's what I would love to see:
1. Flood should read the used/total disk space of all volumes directly from the host, regardless of the host operating system
2. The drives displayed in the sidebar should be configurable in the UI tab of the settings modal (so that a user can hide individual drives as desired)

As for the design itself, I'm still sketching out what that might look like. I'll try to provide a mockup tomorrow or the following day, but I'm open to ideas!

jfurrow:
@RC1140 Here's a quick mockup
<img src="https://cloud.githubusercontent.com/assets/1134755/25261284/1d955042-2606-11e7-86e5-6678276b2325.png" width="240"> <img src="https://cloud.githubusercontent.com/assets/1134755/25261285/1da69f96-2606-11e7-92dd-86f75d7bbc1d.png" width="240">

RC1140:
Wow thanks for that , makes things much easier to build, gonna try and work on it more this weekend been just putting together scafolding atm.

Phlooo:
This mockup is perfect!

A similar bar for CPU load would be cool too :)

wopian:
Considering the remaining disk space got removed entirely by https://github.com/jfurrow/flood/commit/fc2fe936fab2fd0a346a0c5011f09bf9ae2074fe, it's now much needed - as the only way to check atm is by SSHing in or reverting to the parent commit

RC1140:
Whoa , I totally forgot about this work , sorry about the delay , will restart work on this.

Soullivaneuh:
@jfurrow the mock up looks good, but why not putting the disk usage below the speed graph? Make more senses to me.

jesec:
@RC1140 

This feature has been implemented by https://github.com/Flood-UI/flood/pull/808. 

You may close the issue if it is no longer relevant.

PR #808, opened 2019-06-09
Feat/diskuse

sktt:
Here's an initial take on #308 only Linux is supported since that's what I am running. Please advice me on components to use or css classes to employ if you see something as this is the first time i work on this project!

## Description
So checking disk usage through `df` (linux) and pushes through the activity stream if changed. This is done on a 10 second interval while the activity stream is open.

## Related Issue
See #308 

## Motivation and Context
I constantly run out of disk space

## How Has This Been Tested?
Only by running on my Linux machine

image

## Types of changes

## Checklist:


server/services/diskUsageService.js
```diff
@@ -0,0 +1,83 @@
...
+ * `BaseService` nor have any use of the per user API ie. `getSerivce()`
+ */
+const EventEmitter = require('events');
+const util = require('util');
+const exec = util.promisify(require('child_process').exec);
+const diskUsageServiceEvents = require('../constants/diskUsageServiceEvents');
+
+const diskUsage = {
+  linux: () =>
+    exec('df --block-size=1 --type=ext4 --portability | tail -n+2').then(({stdout}) =>
```
sktt: This may not include docker partitions which may be of type `overlay` (seen with `df -T`). I guess I could include those by `grep`-ing with a regex (eg. `grep -E 'ext[34]|overlay'` rather than `df --type=ext4`. Are there other funny fs types on linux to include? I guess linux could also mount ntfs drives too hmm...... :)

jfurrow: This command seems to be targeting very specific volumes, can we make it more robust?

server/services/diskUsageService.js
```diff
@@ -0,0 +1,83 @@
...
+            avail: Number.parseInt(avail, 10),
+            target,
+          };
+        }),
+    ),
+  // win32
+  // darwin
+};
+
+const INTERVAL_UPDATE = 10000;
```
jfurrow: Polling every 10 seconds seems a little aggressive to me, but I don't know how expensive `df` is. What do you think about once every 5 minutes?

server/services/diskUsageService.js
```diff
@@ -0,0 +1,83 @@
...
+const diskUsageServiceEvents = require('../constants/diskUsageServiceEvents');
+
+const diskUsage = {
+  linux: () =>
+    exec('df --block-size=1 --type=ext4 --portability | tail -n+2').then(({stdout}) =>
+      stdout
+        .trim()
+        .split('\n')
+        .map(disk => {
+          const [, /* fs */ size, used, avail /* pcent */, , target] = disk.split(/\s+/);
```
jfurrow: What do the comments mean? Also do you happen to know what the values for indices 0 and 4 represent? It would be nice to define the unused variables, prefixed with underscores (e.g. `_foo`).

server/services/diskUsageService.js
```diff
@@ -0,0 +1,83 @@
...
+        this.listenerCount(diskUsageServiceEvents.DISK_USAGE_CHANGE) === 0 &&
+        event === diskUsageServiceEvents.DISK_USAGE_CHANGE
+      ) {
+        clearInterval(this.updateInterval);
+      }
+    });
+  }
+
+  updateDisks() {
+    return diskUsage[process.platform]().then(disks => {
```
jfurrow: Looks like this is going to throw when `process.platform` is not `linux`. If the platform is unsupported, we should either avoid attaching listeners to this service or handle the unsupported platforms gracefully in the service.

client/src/javascript/components/sidebar/DiskUsage.js
```diff
@@ -0,0 +1,74 @@
...
+            <span>
+              <Size value={d.used} /> Used&nbsp;&nbsp;&nbsp;
+              <Size value={d.avail} /> Free
+            </span>
+          }
+          position="bottom"
+          wrapperClassName="">
+          <div>
+            {d.target}
+            <span style={{float: 'right'}}>{Math.round((100 * d.used) / d.size)}%</span>
```
jfurrow: Let's not use inline styles here. Also let's not use `float`, flexbox should work to acheive the desired layout.

client/src/javascript/components/sidebar/DiskUsage.js
```diff
@@ -0,0 +1,74 @@
...
+    this.setState({disks: DiskUsageStore.getDiskUsage()});
+  }
+
+  getDisks() {
+    return this.state.disks.map(d => (
+      <li key={d.target} className="sidebar-filter__item sidebar__diskusage">
+        <Tooltip
+          content={
+            <span>
+              <Size value={d.used} /> Used&nbsp;&nbsp;&nbsp;
```
jfurrow: Let's use CSS for any desired whitespace rather than whitespace characters

Daiyousei:
@sktt said : "Are there other funny fs types on linux to include? I guess linux could also mount ntfs drives too hmm...... :)"

Well, I already see xfs, which is now the default in redhat/centos
Or maybe nfs, for people using NAS/SAN storage.
Same goes for ZFS, sometime found with remote storage
Personally, I have a "fast seeding" folder in a tmpfs directory

sktt:
Thanks, @Daiyousei your comment makes a lot of sense. The only reason I see to not have people add their own directory to watch is just that. Adding more configuration makes it harder to use. So then we should figure out what the default should be: 
1. to show nothing? 
2. to show all mounted disks?
3. try to be clever and find the mounted disks of currently used paths.

I guess 1. would make most people unaware of the feature. I guess 2. would add alot of useless info. I guess 3 is too error prone and a small reward:effort ratio.

jfurrow:
It would be nice if each user could configure a list of directories. I also like your second idea for the default state (showing all mounted disks).

zuosc:
This feature is very useful to me😂

client/src/javascript/components/sidebar/DiskUsage.js
```diff
@@ -0,0 +1,74 @@
...
+    DiskUsageStore.listen(EventTypes.DISK_USAGE_CHANGE, this.onDiskUsageChange);
+  }
+
+  componentWillUnmount() {
+    DiskUsageStore.unlisten(EventTypes.DISK_USAGE_CHANGE, this.onDiskUsageChange);
+  }
+
+  onDiskUsageChange() {
+    this.setState({disks: DiskUsageStore.getDiskUsage()});
+  }
```
jfurrow: I just merged a change which introduces a new pattern for getting data from stores. Please implement that here as well.

[Here's an example](https://github.com/Flood-UI/flood/blob/master/client/src/javascript/app.js#L94-L106) of how the higher order component can be used. After you implement it, you can just consume the data from the component props.

zuosc:
No one can solve this conflict?

Syc0x00:
I'm using WSL (Windows Subsystem for Linux) so `--type=ext4` leads to no disks being found for me. 

```sh
$ df --type=ext4
df: no file systems processed
```

Suggest you add lxfs and drvfs

```sh
$ df --block-size=1 --type=ext4 --type=lxfs --type=drvfs --portability | tail -n+2
rootfs          499462959104 231522574336  267940384768      47% /
C:              499462959104 231522574336  267940384768      47% /mnt/c
D:              524901412864 429621841920   95279570944      82% /mnt/d
E:             3000574668800 448401506304 2552173162496      15% /mnt/e
```

Here are my filesystems with type if you're curious.

```sh
$ df -T -h
Filesystem     Type   Size  Used Avail Use% Mounted on
rootfs         lxfs   466G  216G  250G  47% /
none           tmpfs  466G  216G  250G  47% /dev
none           tmpfs  466G  216G  250G  47% /run
none           tmpfs  466G  216G  250G  47% /run/lock
none           tmpfs  466G  216G  250G  47% /run/shm
none           tmpfs  466G  216G  250G  47% /run/user
C:             drvfs  466G  216G  250G  47% /mnt/c
D:             drvfs  489G  401G   89G  82% /mnt/d
E:             drvfs  2.8T  418G  2.4T  15% /mnt/e
```

jfurrow:
@sktt Would you like to make the requested changes? If so, we can wait a bit longer, but if not, I'd be happy to do it.

sktt:
Hey sure ill try to find some time tomorrow :D
So we list all fs as default and a config field to filter desired ones?
Im happy if someone wants to deal with windows. Osx could probably run the
same code.

I think i had a memory leak though from spawning df. Maybe it’s node
related. Testing on a later node version right now.

Syc0x00:
Just tried it on osx Mojave 10.14.5 (18F132) and unfortunately `df` isn't quite the same. You have to use the `-T` flag to filter by filesystem type and many of the flags available on linux aren't there or are renamed.

```bash
$ df -T apfs
Filesystem   512-blocks      Used Available Capacity iused               ifree %iused  Mounted on
/dev/disk1s1  976490568 224685624 744109752    24% 1722065 9223372036853053742    0%   /
/dev/disk1s4  976490568   6291496 744109752     1%       4 9223372036854775803    0%   /private/var/vm
```

On top of this versions older than High Sierra (10.13) used HFS as their filesystem so depending on the version of OSX you'll have to run something like `df -T hfs` but I don't have an older system to test this on.

I don't know JS so I can't help implement it but I'm happy to test it.

Daiyousei:
Or why do we do simpler ?

Could we just grab the FS configured as default download directory in rtorrent, and allow the user to add the path they want ?

sktt:
@Syc0x00 is it this the case in portability mode too? ie. using -P or --portability flag?
@Daiyousei was thinking of this but that's actually harder since I guess the right way to do that would be query rtorrent over xmlrpc (which i know nothing about) for that directory.

server/services/diskUsageService.js
sktt: the columns of df output.. that's their short names, originally from df manual pages

Syc0x00:
On top not knowing XMLRPC many people have multiple drives where they keep torrented content. By showing all the drives with `df` they'd be able to know which drive to move new downloads to.

server/services/diskUsageService.js
sktt: i could, but this should only happen while there is an open activity stream. and there should always only be one DiskUsageStorage instance to make sure that df is not run more with more open sockets. I don't think it should have much impact performance. I like to see my disk space clearing up after removing a torrent so it's nice from that pov

Syc0x00:
@sktt Looking at the manpage says no, unfortunately OSX likes to be different.
Screen Shot 2019-07-13 at 11 46 21 PM

server/services/diskUsageService.js
sktt: It may however cause some excess disk usage change events and network traffic and ui updates, maybe i should set the blocksize to 1MiB...

sktt:
@jfurrow did some changes, particularily https://github.com/Flood-UI/flood/pull/808/commits/47e5a5ffc877d31e0ce1cab30b7cafc77a936e24 which is a little bit unrelated but makes lets stops services from emitting events when there's no open connection. 

@Syc0x00 I tried adding macos support with the example output you gave. Maybe it works? maybe you could test running this code and see that it returns an array of object with size,used,available,target fields? https://gist.github.com/sktt/c514c7e7247c3de7bc09b037985252e7

Syc0x00:
@sktt 

```bash
$ node test_node.js
[
  {
    size: 499963170816,
    used: 115019517952,
    avail: 381003714560,
    target: '/'
  },
  { size: 340992, used: 340992, avail: 0, target: '/dev' },
  {
    size: 499963170816,
    used: 3221245952,
    avail: 381003714560,
    target: '/private/var/vm'
  },
  { size: NaN, used: 0, avail: 0, target: '100%' },
  { size: NaN, used: 0, avail: 0, target: '100%' }
]
$ uname -a
Darwin Macintosh 18.6.0 Darwin Kernel Version 18.6.0: Thu Apr 25 23:16:27 PDT 2019; root:xnu-4903.261.4~2/RELEASE_X86_64 x86_64
```

Syc0x00:
Reading the [manpage ](https://ss64.com/osx/df.html) for `df` on OSX shows that you actually can filter for multiple filesystems.

`df -t apfs,hfs`

jfurrow:
> @jfurrow did some changes, particularily 47e5a5f which is a little bit unrelated but stops services from emitting events when there's no open connection. Not sure how this worked before because it looks like it continues to emit server events after the connection has been closed.

Good catch, your change looks right to me.

jfurrow:
> Im happy if someone wants to deal with windows

We can forego Windows support for now, as long as it's handled gracefully. I'll set up a Windows VM to test.

server/services/diskUsageService.js
jfurrow: This is handled now

server/services/diskUsageService.js
jfurrow: I'm still not crazy about the 10 second poll interval, but it seems fine for now. I assume Flood users' sessions are usually pretty short, so running every 5 seconds while their session is active hopefully won't be problematic.

If you're up for it, I think it would be better to use a long poll interval (like 5 minutes), and then manually invoke `diskUsageService.updateDisks` when disk usage is expected to have changed (e.g. after the server handles a request to remove torrents).

If you take this approach, I also suggest modifying this code such that the `updateDisks` invocation is scheduled via `setTimeout`. `updateDisks` would be responsible for clearing and setting its own delayed invocation, so that it's invoked immediately when it needs to be, but never more than once per 5 minutes.

Syc0x00: Along with that idea why not just make it a configurable setting in config.js or in the settings page. That way those who want quick updates have that option avaible to them.

sktt:
maybe you also need the `-l` flag to filter some rows 

Yea I guess there's also the option to rely on a dependecy, but maybe more as a source of inspiration for now. https://github.com/sv-code/microstats/

server/services/diskUsageService.js
ngseer: Here are some filtering criterias used in the Prometheus. Might come in handy.
https://github.com/prometheus/node_exporter/blob/master/collector/filesystem_linux.go#L32

server/services/diskUsageService.js
```diff
@@ -0,0 +1,116 @@
...
+          return {
+            size: Number.parseInt(size, 10) * 1024,
+            used: Number.parseInt(used, 10) * 1024,
+            avail: Number.parseInt(avail, 10) * 1024,
+            target,
+          };
+        }),
+    ),
+  // TODO:
+  win32: () => Promise.resolve([]),
```
jplot: Here is the command for get disks on windows:
`wmic logicaldisk get name, freespace, size`
```FreeSpace     Name  Size
115519344640  C:    201439305728
              D:
425615523840  Y:    1998251409408
26653650944   Z:    298000031744
```

jfurrow:
@sktt how are you feeling about this PR? Do you want to make the requested changes? If not, it's okay!

server/services/diskUsageService.js
sktt: hi @jplot! Can you see how this little snippet runs on windows?

https://gist.github.com/sktt/0b89d0c90ed91825e865cf2ce35732af

@jfurrow i guess this could be added at a later stage though..

Issue #823, opened 2019-09-02
Disk usage not reporting


MattMattV:
Type: Bug Report

- [x] Try to follow the update procedure described in the README and try again before opening this issue.
- [x] Please check the [F.A.Q.](https://github.com/Flood-UI/flood/wiki/F.A.Q.).
- [x] Please check the [Troubleshooting](https://github.com/Flood-UI/flood/wiki/Troubleshooting) wiki section.

## Your Environment
Flood is running in a Docker container using the provided Dockerfile.
* Version used: 86008d5ffb5d073278bf29f46b874b85495980ed
<details>
<summary>Docker engine version</summary>
<pre>
Version:          18.06.1-ce
API version:      1.38 (minimum version 1.12)
Go version:       go1.10.3
Git commit:       e68fc7a
Built:            Tue Aug 21 17:23:21 2018
OS/Arch:          linux/amd64
Experimental:     false
</pre>
</details>

* Operating System and version (inside container):

Alpine Linux 3.9.4

## Summary
Disk widget indicates `NaN%`

![disk widget showing NaN%](https://user-images.githubusercontent.com/6715182/64110385-48a92d00-cd82-11e9-8260-012757c2150d.png)

## Expected Behavior
Disk usage widget reporting space usage

## Current Behavior
`NaN%` is show instead of targeted storage usage

## Possible Solution
Review `df` options to take account of various Linux distributions

## Steps to Reproduce
1. Build docker image at root of repository
1. Run it with a volume to host downloaded data
1. Download torrent

## Appendix

<details>
<summary>Output of the "df" command executed inside the container</summary>
<pre>
/usr/src/app # df --block-size=1024 --portability
df: unrecognized option: block-size=1024
BusyBox v1.29.3 (2019-01-24 07:45:07 UTC) multi-call binary.

Usage: df [-PkmhTai] [-B SIZE] [FILESYSTEM]...

Print filesystem usage statistics

	-P	POSIX output format
	-k	1024-byte blocks (default)
	-m	1M-byte blocks
	-h	Human readable (e.g. 1K 243M 2G)
	-T	Print filesystem type
	-a	Show all filesystems
	-i	Inodes
	-B SIZE	Blocksize
</pre>
</details>

<details>
<summary>Output of "df -B 1024" executed inside the container</summary>
<pre>
/usr/src/app # df -B 1024
Filesystem           1K-blocks      Used Available Use% Mounted on
overlay              1902052484 1527146244 278264552  85% /
[...]

PR #825, opened 2019-09-03
diskusage : remove useless option

noraj:
## Description

see the discussion here https://github.com/Flood-UI/flood/issues/823#issuecomment-527592487

## Related Issue

#823

## Motivation and Context

simplify code

## How Has This Been Tested?

see  https://github.com/Flood-UI/flood/issues/823#issuecomment-527592487

## Types of changes

## Checklist:

### DIFF
--- a/server/services/diskUsageService.js
+++ b/server/services/diskUsageService.js
@@ -19,7 +19,7 @@ const filterMountPoint =
 
 const diskUsage = {
   linux: () =>
-    execFile('df --block-size=1024 --portability | tail -n+2', {
+    execFile('df | tail -n+2', {
       shell: true,
       maxBuffer: 4096,
     }).then(({stdout}) =>


sktt:
Awesome! :D

server/services/diskUsageService.js
```diff
@@ -19,7 +19,7 @@ const filterMountPoint =
 
 const diskUsage = {
   linux: () =>
-    execFile('df --block-size=1024 --portability | tail -n+2', {
+    execFile('df | tail -n+2', {
```
sktt: I think the problem was that apline does not support any of the long names, perhaps we can use `-k -P` instead?

I think `-P` is healthy so that we're sure the columns are in the same order.
I think explicit block-size `-k` could be useful too if some system uses single byte blocksize or otherwise.

noraj: I won't explain as long as in #823 but I think the posix option just standardized the column naming, not the columns order. I also think that all linux will have the same default block size, it may change only on different platform like *BDS or MacOS but any the script only handle linux and mac for now.
I think the options would be useless but if it safer we can use `df -kPl` and so have teh same command used for `['darwin', 'linux'];` simplifying the code. Someone just need to confim that `-P` is also working on Mac.

MacTwister: Had no issues running on MacOS 10.11:
```
➜  ~ df -kPl
Filesystem                        1024-blocks      Used Available Capacity  Mounted on
/dev/disk0s2                        487712924 342782424 144674500    71%    /
localhost:/yxD52ojDj7YEU08xiCrBrk   487712924 487712924         0   100%    /Volumes/MobileBackups
```

PR #829, opened 2019-09-14
Fix diskusage crash on unsupported platforms + add freebsd

inve1:
When using flood on an unsupported platform, the app wasn't loading because of this exception:

```
GET /api/activity-stream?historySnapshot=fiveMin 200 1.189 ms - -
TypeError: diskUsage[process.platform] is not a function
    at DiskUsageService.updateDisks (/usr/home/inve1/flood/server/services/diskUsageService.js:99:39)
    at module.exports (/usr/home/inve1/flood/server/middleware/clientActivityStream.js:46:20)
    at Layer.handle [as handle_request] (/usr/home/inve1/flood/node_modules/express/lib/router/layer.js:95:5)
    at next (/usr/home/inve1/flood/node_modules/express/lib/router/route.js:137:13)
    at module.exports (/usr/home/inve1/flood/server/middleware/eventStream.js:26:3)
    at Layer.handle [as handle_request] (/usr/home/inve1/flood/node_modules/express/lib/router/layer.js:95:5)
    at next (/usr/home/inve1/flood/node_modules/express/lib/router/route.js:137:13)
    at Route.dispatch (/usr/home/inve1/flood/node_modules/express/lib/router/route.js:112:3)
    at Layer.handle [as handle_request] (/usr/home/inve1/flood/node_modules/express/lib/router/layer.js:95:5)
    at /usr/home/inve1/flood/node_modules/express/lib/router/index.js:281:22
```

## Description
I put in a check for supported platforms on that function. Since the freebsd df output is the same as linux I also added it as a supported platform with a small refactor

A note about the `maxBuffer` change: My df output is currently 31045 bytes long. In my case this is because of hundreds of rar2fs mounts to make rar downloads plex-friendly. Perhaps filtering filesystems like these before parsing them in node could be an option since the disk usage from these is useless. Don't want to handle this in a bugfix pr though.

## Motivation and Context
Flood stopped working on freebsd after I pulled master today 

## How Has This Been Tested?
Ran it on freebsd

## Types of changes

## Checklist:

### DIFF
--- a/server/services/diskUsageService.js
+++ b/server/services/diskUsageService.js
@@ -8,7 +8,8 @@ const execFile = util.promisify(require('child_process').execFile);
 const config = require('../../config');
 const diskUsageServiceEvents = require('../constants/diskUsageServiceEvents');
 
-const PLATFORMS_SUPPORTED = ['darwin', 'linux'];
+const PLATFORMS_SUPPORTED = ['darwin', 'linux', 'freebsd'];
+const MAX_BUFFER_SIZE = 65536;
 
 const filterMountPoint =
   config.diskUsageService && config.diskUsageService.watchMountPoints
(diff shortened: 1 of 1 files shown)


server/services/diskUsageService.js
```diff
@@ -8,7 +8,8 @@ const execFile = util.promisify(require('child_process').execFile);
 const config = require('../../config');
 const diskUsageServiceEvents = require('../constants/diskUsageServiceEvents');
 
-const PLATFORMS_SUPPORTED = ['darwin', 'linux'];
+const PLATFORMS_SUPPORTED = ['darwin', 'linux', 'freebsd'];
```
noraj:
```diff
-freebsd
+bsd
```

As it is Linux and not a distro name.

inve1: Sure but process.platform returns the string ‘freebsd’. Also I wouldn’t assume the output of df is consistent across all BSDs. technically Darwin is a BSD and the output is different :P

noraj: @inve1 You are totally right, my bad.

noraj:
LGTM : Looks Good To Me (I had to search lol)

Issue #739, opened 2018-11-12
Docker: Improvement suggestions


bootldrDNB:
## Permissions
The current Dockerfile is currently running Flood as root.
This is a security issue as [containers shouldn't be run as root](https://medium.com/@mccode/processes-in-containers-should-not-run-as-root-2feae3f0df3b) and a problem for rTorrent as it is unable to write to the directories that Flood creates without using setting `chmod 777` on the target folder.

A way to fix this would be to use [s6-overlay](https://github.com/just-containers/s6-overlay) and specify the UID and GID via an environment variable, create a run script which then runs Flood with the variables you've set, or specify a user within the Dockerfile, like they do in [this example](https://github.com/nodejs/docker-node/blob/master/docs/BestPractices.md#non-root-user).

## Configuration
After speaking to a user in the Flood Discord server, I found out that we both hard-code values in `config.docker.js` because environment values aren't being changed in the configuration file while the container is being built. 

The environment values are being passed to the container - _you can echo `$FLOOD_BASE_URL` in the container shell  and you'll receive the value you passed in your `docker-compose.yml` config file_, however the configuration file is never modified to use the values passed by the environment variables.

## Docker wiki not being clear enough
One thing I noticed after following the guide was that I couldn't move files at all using Flood, because Flood didn't know of the download directory I used.

However, I found a solution in #581 for this issue, but I wish this would've been mentioned in the wiki.

---
_Disclaimer_; I'm very new to Docker, so I would like to apologize if I use the wrong terminology for things. If you're open to PRs, I could possibly try to figure out a solution for these problems and create a pull request. Would be nice to contribute to a open source project I really enjoy otherwise!

noraj:
Hi @bootldrDNB thank for reporting this issue. I usually close issues not respecting the template but I'm going to make an exception for the sake of simplicity (avoiding you to create 3 separate issues).

jfurrow:
@bootldrDNB I'm not using the docker image in this repo — it was submitted by members of the community who wanted it. If you're up for it, it would be wise for you to submit a PR with these changes.

Issue #621, opened 2018-02-14
FLOOD_BASE_URI does not work in Docker image


sgsunder:
I am trying to run flood, using the provided Dockerfile, behind a nginx reverse proxy.

In my docker-compose file I pass the environment variable `FLOOD_BASE_URI = /flood` to the container. I have even checked using `docker exec` to make sure this variable is actually passed.

When accessing `http://192.168.1.200/flood` from a browser returns a blank, dark blue page. 

Viewing the source of the page reveals the problem: 
`<script src="/static/js/main.3cee5571.js" type="text/javascript"></script>`

Of course, the URL should be `/flood/static/js/main.3cee5571.js`, so it seems that flood is not properly using the baseURI variable in generating the page.

Testing further:
- Running `wget http://192.168.1.200/static/js/main.3cee5571.js` returns a 404 error, as expected
- Running `wget http://192.168.1.200/flood/static/js/main.3cee5571.js` works and returns a javascript file. So this seems to indicate to me that nginx is not the issue, as it is forwarding URLs correctly when properly formatted

Surprisingly, bypassing the proxy by accessing `http://192.168.1.200:3000/` still gives a fully functional flood page, even with the baseURI variable set to '/flood', and `http://192.168.1.200:3000/flood` still does not work. Again, this seems to me an issue with flood not handling the baseURI variable properly, instead of an issue with the reverse proxy configuration.

Here's my nginx proxy configuration
```   
location /flood/ {
        proxy_http_version 1.1;
        rewrite ^/flood/(.*) /$1 break;
        proxy_pass http://localhost:10660/flood/;
        proxy_set_header  Host            $http_host;
        proxy_set_header  Upgrade         $http_upgrade;
        proxy_set_header  Connection      "upgrade";
        proxy_set_header  X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header  X-Scheme        $scheme;
        proxy_set_header  X-Script-Name   /flood;
}
```

I also tried running without the `proxy_set_header` commands with no luck.

## Your Environment

* Version used: Dockerfile built from latest commit (2cb40c4), NGINX 1.13
* Environment name and version (Firefox version 58 on macOS 10.12, MS Edge 41 on Win10, Safari 11.0.3 on macOS 10.12):
* Operating System and version: (Debian Stretch running Docker version 17.12.0-ce)

deed02392:
[...]

PR #837, opened 2019-09-25
docker: use node lts and run the app as flood user

mycodeself:
Changes to the dockerfile.

## Description
The dockerfile has been updated with some improvements:
1. The node version was downgrade to the LTS version of node 10.16.
2. Add and argument to set the FLOOD_BASE_URI configuration option in build time with `docker build --build-arg FLOOD_BASE_URI=/flood/ .`.
3. Create new user for run the application, creating new group and user with name `flood`. The /data directory and the $WORKDIR are owned by this user.

## Related Issue
#739 
#621 

## Motivation and Context

- This change allow the users to change the FLOOD_BASE_URI without edit the config.docker.js file.
- Avoiding the use of the root user is a good practice  

## How Has This Been Tested?

1. Run `docker build --build-arg FLOOD_BASE_URI=/flood/ -t flood . && docker run -p 3000:3000 --name flood flood`
2. Go to browser and view `http://127.0.0.1:3000/flood/`

## Types of changes

## Checklist:


jfurrow:
Thanks @mycodeself!

Dockerfile
```diff
@@ -1,20 +1,23 @@
...
 # Generate node_modules
 COPY package.json \
-     package-lock.json \
-     .babelrc \
-     .eslintrc.js \
-     .eslintignore \
-     .prettierrc \
-     ABOUT.md \
-     $WORKDIR
+    package-lock.json \
```
jfurrow: nit: Can we restore the previous, unless there's a good reason to change it?

Dockerfile
```diff
@@ -1,20 +1,23 @@
-ARG NODE_IMAGE=node:12.2-alpine
+ARG NODE_IMAGE=node:10.16-alpine
```
jfurrow: Looks like v12 will be LTS in a month, do you really think this is worth changing? https://nodejs.org/en/about/releases/

MacTwister: Yeah agree, I would not suggest downgrading the Node version if all the code has been testing/working with v12 plus it's still the "current" version of NodeJS.

Dockerfile
```diff
@@ -39,10 +42,19 @@ WORKDIR $WORKDIR
...
+# Add user to run the application
+RUN addgroup -S flood
+RUN adduser -S flood -G flood
+RUN mkdir /data
+RUN chown flood:flood /data
+RUN chown flood:flood ${WORKDIR}
+
+USER flood
+
 COPY --from=nodebuild $WORKDIR $WORKDIR
```
MacTwister: Docker Docs:

> The `USER` instruction sets the user name (or UID) and optionally the user group (or GID) to use when running the image and for any `RUN`, `CMD` and `ENTRYPOINT` instructions that follow it in the Dockerfile.

The `COPY/ADD` instructions creates new files and directories with a UID and GID of 0.

Therefor if we are going to run the app with the `flood` user, I think, we should also copy the files with `flood` owership. There is a optional `--chown` flag to copy using a specified ower, since Docker 17.09.

```suggestion
COPY --from=nodebuild --chown=flood:flood $WORKDIR $WORKDIR
```

Dockerfile
```diff
@@ -1,20 +1,23 @@
-ARG NODE_IMAGE=node:12.2-alpine
+ARG NODE_IMAGE=node:10.16-alpine
 ARG WORKDIR=/usr/src/app/
+ARG FLOOD_BASE_URI=/
```
MacTwister: I was interested in this change myself, however the static asset files did not compile with the new URI when I build the image.

The `ARG` needs to be defined in a `FROM` block to be used with-in-it, so I suggest moving this line 3 to after line 6 as the argument is only needed in the nodebuild stage build.

So change could look like:
```Dockerfile
FROM ${NODE_IMAGE} as nodebuild
ARG WORKDIR

ARG FLOOD_BASE_URI=/
ENV FLOOD_BASE_URI $FLOOD_BASE_URI
```

[Docker docs](https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact):
> An ARG declared before a FROM is outside of a build stage, so it can’t be used in any instruction after a FROM. To use the default value of an ARG declared before the first FROM use an ARG instruction without a value inside of a build stage.

Dockerfile
mycodeself: I agree with you, i did not know the life cycle of node.js. I will setup the curent stable `ARG NODE_IMAGE=node:12.11-alpine`

Dockerfile
```diff
@@ -39,10 +42,19 @@ WORKDIR $WORKDIR
 RUN apk --no-cache add \
     mediainfo
 
+# Add user to run the application
+RUN addgroup -S flood
+RUN adduser -S flood -G flood
+RUN mkdir /data
+RUN chown flood:flood /data
```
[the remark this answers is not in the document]
mycodeself: Have you been able to test how it affects current configurations?

Dockerfile
mycodeself: I will add this suggestion. Then i think that `RUN chown flood:flood /data` can be removed.

Dockerfile
mycodeself: The previous indentation has 5 spaces. The rest of file has 4 spaces, i think it should use the same. What do you think? If you want to mantain 5 spaces in these lines, i can revert it.

Dockerfile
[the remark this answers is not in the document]
noraj: @mactwister maybe the file are own by root because the docker image was built and the default user was root but if you build it and add another user the file will be own by another user.

Dockerfile
```diff
@@ -1,20 +1,24 @@
-ARG NODE_IMAGE=node:12.2-alpine
+ARG NODE_IMAGE=node:12.11-alpine
```
MacTwister: So turns there is no tag for v12.11 of NodeJS, yet, as there seems to be an issue on NoeJS/Docker side for building the docker images. See pending pull request https://github.com/nodejs/docker-node/pull/1113

However, what I suggest is using the major version only tag as the base of this image. Doing so also saves us having to update our `Dockerfile` for every minor version released. Meaning every time one builds the docker image for Flood, the latest version of node will be used as the base (if you pull while building `docker build --pull` option. Minor version theoretically not have any BC and if anything does stop working, that should warrant a fix in the app anyway as someone maybe running that version.

```suggestion
ARG NODE_IMAGE=node:12-alpine
```

Dockerfile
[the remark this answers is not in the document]
MacTwister: Yes, so now if you build the docker image on your side using this Dockerfile (with chown changes) then the app will be owned and run as the `flood` user. 👍 

#### For new setups using docker:
New container will run as `flood` user and create the database under `/data` owned by the `flood` user. All good here.

#### For existing setups using docker:

The scenario I am describing is for people already running with Docker and have setup the `/data` folder as a "volume" to persist the DB when restarting/upgrading the docker image. Below is simplified version to how I have my setup:

```docker-compose.yml
# ... rest of you docker compose file
  floodui:
    build: .
    ports:
      - 3000:3000
    volumes:
      - flood_data:/data
```

**So** when re-creating the container with new image, the app will crash with the following error:

```
[Error: EACCES: permission denied, open '/data/server/db/users.db'] {
```

This is because my volume `/data` is persisted over from the previous container, where it was first created (when container was root only) so my version of `/data` is owned by `root:root`.

```bash
julian$ docker-compose run --rm floodui ls -l /data
  total 4
  drwxr-xr-x    3 root     root          4096 Oct  3 08:41 server
```

**What I am saying** is that, if we go through with this change, we'd need to some how inform users what they would need to do to continue working after this upgrade. i.e. update the docs for docker setup to inform the to update the /data ownership manually.

#### What I needed to do using my docker compose setup to continue working:
```bash
julian$ docker-compose run --user=root --rm floodui chown -R flood:flood /data
```

Fleshgrinder:
There is not really a need for this because the container can be started with a different user and group and everything works just fine. I would never have used flood if it would require root privileges. Simply use the `--user` option in `docker run` or the `user:` option in `docker-compose.yaml`.

PR #838, opened 2019-09-26
Disk Usage GUI

tillkrischer:
## Description
GUI configuration for disk usage

## Types of changes

## Checklist:

- [ ] My code follows the code style of this project.

### DIFF
--- a/client/src/javascript/components/modals/settings-modal/DiskUsageTab.js
+++ b/client/src/javascript/components/modals/settings-modal/DiskUsageTab.js
@@ -16,53 +16,51 @@ class DiskUsageTab extends SettingsTab {
   };
 
   componentWillMount() {
-    // assemble disk items from saved "mountPoints" and list of disks "disks"
     const mountPoints = SettingsStore.getFloodSettings('mountPoints');
-    const disks = DiskUsageStore.getDiskUsage().reduce((a, c) => {
-      a[c.target] = c;
-      return a;
+    const disks = DiskUsageStore.getDiskUsage().reduce((disksByTarget, disk) => {
+      disksByTarget[disk.target] = disk;
+      return disksByTarget;
     }, {});
-    const diskItems = [];
+
+    // assemble disk items from saved "mountPoints" and list of disks "disks"
     // first targets saved in mountPoints that exist in disks
-    diskItems.push(
-      ...mountPoints.filter(target => target in disks).map(target => ({
+    // then remaing targets from disks
+    const diskItems = mountPoints
+      .filter(target => target in disks)
+      .map(target => ({
         id: target,
         visible: true,
-      })),
-    );
-    // then remaing targets from disks
-    diskItems.push(
-      ...Object.keys(disks)
-        .filter(target => !mountPoints.includes(target))
-        .map(target => ({
-          id: target,
-          visible: false,
-        })),
-    );
+      }))
+      .concat(
+        Object.keys(disks)
+          .filter(target => !mountPoints.includes(target))
+          .map(target => ({
+            id: target,
+            visible: false,
+          })),
+      );
+
     this.setState({diskItems});
   }
 
-  updateSettings = () => {
-    const {diskItems} = this.state;
+  updateSettings = diskItems => {
     const mountPoints = diskItems.filter(item => item.visible).map(item => item.id);
     this.props.onSettingsChange({mountPoints});
   };
 
   handleDiskCheckboxValueChange = (id, value) => {
-    let {diskItems} = this.state;
+    const {diskItems} = this.state;
 
-    diskItems = diskItems.map(disk => {
+    const newItems = diskItems.map(disk => {
       if (disk.id === id) {
-        disk.visible = value;
+        return {...disk, visible: value};
       }
       return disk;
     });
 
-    this.setState({diskItems});
-    this.updateSettings();
+    this.setState({diskItems: newItems});
+    this.updateSettings(newItems);
   };
-
-  handleFormChange = ({event, formData}) => {};
 
   handleDiskMouseDown = () => {
     if (this.tooltipRef != null) {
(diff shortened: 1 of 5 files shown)


client/src/javascript/i18n/en.js
```diff
@@ -188,6 +188,7 @@ export default {
   'settings.tabs.resources': 'Resources',
   'settings.tabs.authentication': 'Authentication',
   'settings.tabs.userinterface': 'User Interface',
+  'settings.tabs.diskuage': 'Disk Usage',
```
sktt: typo :)

client/src/javascript/components/modals/settings-modal/DiskUsageTab.js
```diff
@@ -0,0 +1,134 @@
...
+  };
+
+  componentWillMount() {
+    // assemble disk items from saved "mountPoints" and list of disks "disks"
+    const mountPoints = SettingsStore.getFloodSettings('mountPoints');
+    const disks = DiskUsageStore.getDiskUsage().reduce((a, c) => {
+      a[c.target] = c;
+      return a;
+    }, {});
+    const diskItems = [];
```
sktt: I think think this could be simplified with a single assignment, for example
```js
const diskItems = DiskUsageStore.getDiskUsage().map(d => ({ 
  id: d.target, 
  visible: mountPoints.includes(d.target)
}))
```

tillkrischer: The selected items have to in the order that they appear in `mountPoints` for the reorder functionality to work. That's why I am doing it in this convoluted way.

tillkrischer: The `SortableList` component expects the `visible` key. I looked into changing it, but the User Interface Tab and the `torrentDetails` field in the database use `visible` already.

client/src/javascript/components/modals/settings-modal/DiskUsageTab.js
```diff
@@ -0,0 +1,134 @@
...
+    diskItems: [],
+  };
+
+  componentWillMount() {
+    // assemble disk items from saved "mountPoints" and list of disks "disks"
+    const mountPoints = SettingsStore.getFloodSettings('mountPoints');
+    const disks = DiskUsageStore.getDiskUsage().reduce((a, c) => {
+      a[c.target] = c;
+      return a;
+    }, {});
```
jfurrow: Let's rename variables `a` and `c` to something more meaningful. For example, `c` would be better represented as `disk` (or whatever it actually represents), and `a` would be better represented as `diskByTarget` or something like that.

client/src/javascript/components/modals/settings-modal/DiskUsageTab.js
```diff
@@ -0,0 +1,134 @@
...
+    );
+    // then remaing targets from disks
+    diskItems.push(
+      ...Object.keys(disks)
+        .filter(target => !mountPoints.includes(target))
+        .map(target => ({
+          id: target,
+          visible: false,
+        })),
+    );
```
jfurrow: I have a stylistic nit, and for posterity I'll document it. But don't feel obligated to implement it as it's trivial:

I would try to chain this, maybe something like:
```js
const diskItems = mountPoints
  .filter(target => target in disks)
  .map(target => ({
    id: target,
    visible: true,
  }))
  .concat(
    Object.keys(disks)
      .filter(target => !mountPoints.includes(target))
      .map(target => ({
        id: target,
        visible: false,
      }))
  );
```

client/src/javascript/components/modals/settings-modal/DiskUsageTab.js
```diff
@@ -0,0 +1,134 @@
...
+    let {diskItems} = this.state;
+
+    diskItems = diskItems.map(disk => {
+      if (disk.id === id) {
+        disk.visible = value;
+      }
+      return disk;
+    });
+
+    this.setState({diskItems});
```
jfurrow: Let's rewrite this as a functional `setState`, and avoid mutating `disk`:
```js
this.setState(state => {
  return {
    ...state,
    diskItems: diskItems.map(disk => {
      if (disk.id === id) {
        return {...disk, visible: value};
      }
      return disk;
    })
  }
});
```

client/src/javascript/components/modals/settings-modal/DiskUsageTab.js
```diff
@@ -0,0 +1,134 @@
...
+    this.updateSettings();
+  };
+
+  handleFormChange = ({event, formData}) => {};
```
jfurrow: Do we need this at all?

client/src/javascript/components/sidebar/DiskUsage.js
```diff
@@ -19,37 +20,45 @@ const DiskUsageTooltipItem = ({label, value}) => {
...
-            </div>
-            <ProgressBar percent={(100 * d.used) / d.size} />
-          </Tooltip>
-        </li>
-      );
-    });
+    const {disks, mountPoints} = this.props;
+    const diskMap = disks.reduce((a, c) => {
+      a[c.target] = c;
+      return a;
```
jfurrow: Again here, let's be more verbose about what `a` and `c` represent

client/src/javascript/components/modals/settings-modal/DiskUsageTab.js
jfurrow: Oh, probably because `SettingsTab` calls it. Sorry, I've forgotten what atrocious code I've written

tillkrischer: It seems to work without it.

client/src/javascript/components/modals/settings-modal/DiskUsageTab.js
tillkrischer: My code was actually broken and only working because I was mutating the state directly. 
I fix it now and am not mutating anything.

PR #842, opened 2019-09-30
Unescape Text nodes when deserializing

sktt:
Downloading a finished torrent sometimes fails when any of the file paths contain an ampersand (&), it appears that the retrieved files from the system.multicall in `ClientRequest#getTorrentDetails` have escaped "&" to "&amp;". It appears to be the same with <, >, ", ' too.

## Description
I replace "&amp" with "&" on file paths, which fixes the bug on my end.
I also replace, &{lt,gt,apos,quot}; which should cover all escaped text node content according to these sources:
https://www.w3.org/TR/xml/#syntax
https://stackoverflow.com/questions/1091945/what-characters-do-i-need-to-escape-in-xml-documents

## How Has This Been Tested?
Tested on rtorrent 0.9.6

## Types of changes

## Checklist:

### DIFF
--- a/server/util/rTorrentDeserializer.js
+++ b/server/util/rTorrentDeserializer.js
@@ -1,5 +1,13 @@
 const saxen = require('saxen');
 
+const unescapeXMLString = value =>
+  value
+    .replace(/&amp;/g, '&')
+    .replace(/&lt;/g, '<')
+    .replace(/&gt;/g, '>')
+    .replace(/&apos;/g, "'")
+    .replace(/&quot;/g, '"');
+
 let stackMarks;
 let dataStack;
 let tmpData;
@@ -16,7 +24,7 @@ const openTag = elementName => {
 };
 
 const onText = value => {
-  tmpData.push(value);
+  tmpData.push(unescapeXMLString(value));
 };
 
 const onError = err => {


sktt:
Actually could it be my rtorrent configuration that is bad? also torrents with '&' in the name will show up in the list with "&amp;" instead.
image

server/util/clientResponseUtil.js
```diff
@@ -107,7 +107,11 @@ const processTorrentDetails = data => {
   }
 
   if (filesData && filesData.length) {
-    files = mapPropsToResponse(torrentFilePropsMap.props, filesData);
+    files = mapPropsToResponse(torrentFilePropsMap.props, filesData).map(file => ({
+      ...file,
+      path: file.path.replace(/&amp;/g, '&'),
+      pathComponents: file.pathComponents.map(c => c.replace(/&amp;/g, '&')),
+    }));
```
sktt: Perhaps this should be done in torrentFilePropsMap instead?

jfurrow:
Honestly I'm not sure how Flood handles strings today. I wouldn't be surprised if this issue is consistently reproduced, but I haven't experienced it myself. I'll try to make time to look into it, but if somebody else could, that would be great.

I will say the one-off character replacement in this PR is probably insufficient — we'll need to ensure proper encoding/decoding of utf-8 strings, if Flood isn't already handling them properly.

sktt:
@jfurrow Sorry about that, I've updated my pr description. I check for the five escape characters listed here https://stackoverflow.com/questions/1091945/what-characters-do-i-need-to-escape-in-xml-documents
I think this should cover everything? Other utf8 characters do not need to be escaped so it should be fine. At least some russian seems to work just fine: 
image

jesec:
Merged by https://github.com/jesec/flood/commit/47b2053ac3a6ce78ba57f9396b3487be69877b63.

PR #880, opened 2020-07-29
Add an option to completely disable users and authentication

### LINKED ISSUE #240: Make authentication optional
I know this is well rooted in your app but I would very much like to disable auth / passport and use basic auth on my reverse proxy instead.

Thanks!

nitnelave:
# Description
This change adds a config option `disableUsersAndAuth`. When it is enabled, the app makes no effort to try to authenticate users, skips the login and register pages and goes straight to the dashboard, using a hardcoded administrator user.

The settings for the hardcoded user are in the config under the block `configUser`.

## Related Issue
Fixes #240.

## Motivation and Context
Some servers have their own authentication methods (e.g. nginx basic_auth, LDAP, OAuth and so on). Instead of trying to re-implement everything here, make authentication optional so it can be added as a layer.

## How Has This Been Tested?
I set it up on my server, tested it with rTorrent both with SCGI and socket settings. Monitoring the network requests shows no remaining 401s.
As long as the config flag is not flipped (and by default it's not, which is why I chose the "disable" version instead of "enable" as I would have liked), this change is a no-op.
Note that I didn't try every operation, but given the horizontal nature of the change, I would expect that if something works, everything works.

## Types of changes
- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [?] My code follows the code style of this project.
  I think so, but please double-check!
  I think mentioning it in the Wiki could be nice, but I updated the config template.
- [?] I have read the **CONTRIBUTING** document.
  Couldn't find one :/
  I'll need help on that one, I'm not very comfortable with Node.
  `npm test` -> "There are no tests yet."

### DIFF
--- a/config.docker.js
+++ b/config.docker.js
@@ -15,6 +15,7 @@ const CONFIG = {
   ssl: process.env.FLOOD_ENABLE_SSL === 'true' || process.env.FLOOD_ENABLE_SSL === true,
   sslKey: '/data/flood_ssl.key',
   sslCert: '/data/flood_ssl.cert',
+  disableUsersAndAuth: false,
 };
 
 module.exports = CONFIG;
--- a/config.template.js
+++ b/config.template.js
@@ -19,6 +19,23 @@ const CONFIG = {
   dbCleanInterval: 1000 * 60 * 60,
   // Where to store the local nedb database.
   dbPath: './server/db/',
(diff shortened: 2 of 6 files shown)


jesec:
Merged: https://github.com/jesec/flood/commit/06e6d1e3f7bf2aef068d256b0c7a6dd1280567a0.

